PR #4, opened 2016-12-21
Add basic setuptools and Debian packaging

pwithnall:
Various commits which pythonify things so setuptools is happier; add setuptools; then add Debian packaging based on that.

The package will not currently build successfully, due to a missing python-pyndn dependency. I need to package that before this can be merged. However, I have tested the build with that dependency commented out and pyndn installed through pip, and things seem to work, so this set of patches should not be too far from correct.


eos_data_distribution/DirTools/test_class.py
```diff
@@ -1,5 +1,5 @@
 import pytest
-import DirTools
+import eos_data_distribution.DirTools
```
magcius: This needs to be `from eos_data_distribution import DirTools`, or, you can use `from .. import DirTools`

demo/demo.sh
```diff
@@ -25,8 +25,8 @@ run () {
         echo "$title '$t'; $(cprofile ${BASE_PATH}/$@) 2>&1 | tee ${BASE_PATH}/$t.log; sleep infinity"
 }
 
-run_router="$(run router soma-subscriptions-producer)"
```
magcius: Please don't use `__init__.py` -- that means this code is parsed and loaded upon any load of the package, and any code in `__init__.py` other than imports is considered poor form by experienced Python users.

Please use `eosdatadistribution/tools/ostree_store.py` or so. Also, use `python -m eosdatadistribution.foo` instead of running scripts directly so that the search path is set up correctly.

debian/changelog
```diff
@@ -0,0 +1,12 @@
+eos-data-distribution (0.1.0) UNRELEASED; urgency=medium
```
magcius: Please put debian-specific stuff onto the `debian-master` branch. Our CI system knows how to merge `debian-master` and `master` together when building.

eosdatadistribution/DirTools/__init__.py
```diff
@@ -5,7 +5,7 @@
 
 
 class Monitor(GObject.GObject):
-    __gsignals__ = {v: (GObject.SIGNAL_RUN_FIRST, None, (str, object, object, object, object, object)) for k, v in EventToSignal.items()}
+    __gsignals__ = {v: (GObject.SIGNAL_RUN_FIRST, None, (str, object, object, object, object, object)) for k, v in list(EventToSignal.items())}
```
magcius: This is already fine, no need to convert to list.

eosdatadistribution/MDNS/__init__.py
```diff
@@ -39,7 +39,7 @@ def __init__(self):
         self.pretty_name = avahi.ServiceTypeDatabase.ServiceTypeDatabase()
 
     def get_human_type(self, type):
-        if self.pretty_name.has_key(type):
+        if type in self.pretty_name:
```
magcius: Are we sure `ServiceTypeDatabase` supports `__contains__`?

dbus-service.py
```diff
@@ -92,7 +92,7 @@ def impl_DownloadSubscription(self, invocation, parameters):
         subscription_id, = parameters.unpack()
```
magcius: Most of the code already supported Python 3 -- fixing up the remaining prints should be enough without the other stuff.

debian/changelog
pwithnall: Out of interest, why keep `master` and `debian-master` separate? This is our project: we know that it’s going to be built into a Debian package. Might as well save some rebase and branching faff and keep `debian/` in `master`, right?

demo/demo.sh
pwithnall: > Please use `eosdatadistribution/tools/ostree_store.py` or so. Also, use `python -m eosdatadistribution.foo` instead of running scripts directly so that the search path is set up correctly.

In this case we can’t use `-m`, since it’s already being used: `demo.sh` actually runs `python -m cProfile -o eosdatadistribution/foo.prof eosdatadistribution/foo.py`, and cprofile appears to only take Python filenames as arguments.

eosdatadistribution/MDNS/__init__.py
pwithnall: [Seems to](https://github.com/lathiat/avahi/blob/master/avahi-python/avahi/ServiceTypeDatabase.py.in#L81)

debian/changelog
pwithnall: (If you still want this done, I’ll split out the commit before merging, to avoid having too many pending branches.)

magcius: We try to treat each codebase we build as an "upstream" project, and upstream projects usually do not have debian directories.

demo/demo.sh
magcius: Then the usual way to do this is to add an `entry_point` script using distutils / setuptools and using the entrypoint script your argument to `cProfile`.

eosdatadistribution/MDNS/__init__.py
magcius: Please split this out then -- this isn't part of a Python 2 to Python 3 port, 2to3 is just doing a buggy find/replace.

demo/demo.sh
pwithnall: In the interests of getting the massive rename commit in before things pile up behind it, can the ‘do this right’ change be deferred to a follow-up commit here?

debian/changelog
pwithnall: hmm, I think that’s a bit pointless when there are realistically no other consumers of a project (it’s closed source, even), but whatever. I’ll split this into a separate branch when merging.

eosdatadistribution/MDNS/__init__.py
pwithnall: OK.

debian/changelog
magcius: It's not closed source though?

demo/demo.sh
magcius: Sure, that's fine.

COPYING
```diff
@@ -0,0 +1,165 @@
+                   GNU LESSER GENERAL PUBLIC LICENSE
+                       Version 3, 29 June 2007
```
xaiki: @cosimoc what's the take on this ? do we want our code to be reuseable in non-free software ?
as i'm mostly the author i'd rather it be gplv3

xaiki: at least the deamons shoud be, perhaps we need a deamons/lib slpit ?

dbus-service.py
```diff
@@ -30,8 +30,8 @@
 
 from pyndn import Name, Face
 
-from eos_data_distribution.ndn import Consumer
-from eos_data_distribution.names import SUBSCRIPTIONS_INSTALLED
+from eosdatadistribution.ndn import Consumer
+from eosdatadistribution.names import SUBSCRIPTIONS_INSTALLED
```
xaiki: i'd really like to avoid this rename...

eosdatadistribution/producers/soma_subscriptions.py
```diff
@@ -25,10 +25,15 @@
 
 from gi.repository import GLib
 
-from eos_data_distribution.soma_subscription_fetcher import Fetcher
+from eosdatadistribution.soma_subscription_fetcher import Fetcher
```
xaiki: so our naming scheme is a bit of the oposite, i.e. $what.py -> {producer,consumer} it usually makes more sense.

eosdatadistribution/producers/usb.py
```diff
@@ -56,7 +56,7 @@ def mount_removed_cb(monitor, mount, store):
     pprint.pprint([store.remove_name(n) for p, n in store.producers.items() if p.startswith(p)])
 
 
-if __name__ == '__main__':
+def main():
```
xaiki: is that needed ? seems to work fine without it, and i feel it makes the code easier to read not to add another brain dereference.

COPYING
pwithnall: I just copied the licensing header from the existing files. If we want to relicense, that’s fine, but let’s do it as a separate commit to reduce the risk of confusion.

dbus-service.py
pwithnall: As discussed on IRC, as far as I know that’s not possible: Python prefers names without underscores; Debian prohibits them in package names; `pybuild` makes it painful to do anything except have the Python and Debian names equal.

eosdatadistribution/producers/soma_subscriptions.py
pwithnall: Fine.

eosdatadistribution/producers/usb.py
pwithnall: I thought it was needed for the `entry_points` stuff in `setup.py`, but I admit I cargo-culted it from another module and haven’t tested it without here.

pwithnall: I’ve re-read on why the cargo-culting was correct, and having a `main()` method is the [recommended approach](https://packaging.python.org/distributing/#scripts) because it is [more testable](http://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-console-scripts-entry-point) — the script can be tested by calling the `main()` function from a unit test, without needing to spawn the script itself.

eosdatadistribution/producers/soma_subscriptions.py
pwithnall: Done.

pwithnall:
Rebased on master, and all non-deferred review comments fixed, as far as I can tell.

Things still to do in follow-up commits:
 * [Fix the way cpython calls the scripts](https://github.com/endlessm/endless-ndn/pull/4#discussion_r93542063)
 * [Fix the copyrights and potentially relicense](https://github.com/endlessm/endless-ndn/pull/4#discussion_r93696153)

And then further commits will add things like systemd `.service` files, etc.

PR #5, opened 2016-12-22
Xaiki/re request

xaiki:
this moves a bunch of things around but really goes towards the goal of making re-request work.

so re-request works by having the date in the name for manifest.json, we do this by supperclassing the http getter and splitting the http only function into another class (http.Getter)

still WIP.

### DIFF
--- a/eos_data_distribution/ndn/base.py
+++ b/eos_data_distribution/ndn/base.py
@@ -30,7 +30,7 @@ from gi.repository import GLib
 from pyndn.node import Node
 from pyndn.security import KeyChain
 from pyndn.transport.unix_transport import UnixTransport
-from pyndn import Name, Data, Face, MetaInfo
+from pyndn import Name, Data, Face
 
 logger = logging.getLogger(__name__)
 
--- a/eos_data_distribution/ndn/base.py
+++ b/eos_data_distribution/ndn/base.py
@@ -203,13 +203,13 @@ class Consumer(Base):
         self._callbackCount += 1
         self.emit('data', interest, data)
 
-    def expressInterest(self, name=None, forever=False):
-        if name is None:
-            name = self.name
+    def expressInterest(self, interest=None, forever=False):
+        if interest is None:
+            interest = self.name
 
-        logger.debug("Express Interest name: %s", name)
-        onTimeout = partial(self.onTimeout, forever=forever, name=name)
-        self.pit[name] = self.face.expressInterest(name, self._onData, onTimeout)
+        logger.debug("Express Interest name: %s", interest)
+        onTimeout = partial(self.onTimeout, forever=forever, name=self.name)
+        self.pit[interest] = self.face.expressInterest(interest, self._onData, onTimeout)
 
     def removePendingInterest(self, name):
         self.face.removePendingInterest(self.pit[name])
--- a/eos_data_distribution/ndn/base.py
+++ b/eos_data_distribution/ndn/base.py
@@ -164,20 +164,12 @@ class Producer(Base):
             logger.warning("tried to unregister a prefix that never was registred: %s", prefix)
             pass
 
-    def registerPrefix(self, prefix=None, postfix="", flags=None):
-        prefix = makeName(prefix)
-        postfix = makeName(postfix)
-
-        if not prefix:
-            prefix = Name(self.name).append(postfix)
-        try:
-            flags = flags or self.flags
-        except:
-            flags = None
+    def registerPrefix(self, prefix=None, flags=None):
+        if prefix is None:
+            prefix = self.name
 
         logger.info("Register prefix: %s", prefix)
         self._prefixes[prefix] = self.face.registerPrefix(prefix, self._onInterest, self.onRegisterFailed, self.onRegisterSuccess, flags=flags)
-        return prefix
 
     def onRegisterFailed(self, prefix):
         self._responseCount += 1
@@ -213,17 +205,13 @@ class Consumer(Base):
         # to call into buf() to get a bytearray…
         self.emit('data', interest, data)
 
-    def makeInterest(self, name):
-        return Name(name)
+    def expressInterest(self, name=None, forever=False):
+        if name is None:
+            name = self.name
 
-    def expressInterest(self, name=None, forever=False, postfix=None):
-        if name == None: name = self.name
-        segname = self.makeInterest(name)
-        if postfix: segname.append(postfix)
-        logger.debug("Express Interest name: %s", segname)
+        logger.debug("Express Interest name: %s", name)
         onTimeout = partial(self.onTimeout, forever=forever, name=name)
-        self.pit[name] = self.face.expressInterest(segname, self._onData, onTimeout)
-        return segname
+        self.pit[name] = self.face.expressInterest(name, self._onData, onTimeout)
 
     def removePendingInterest(self, name):
         self.face.removePendingInterest(self.pit[name])
--- a/eos_data_distribution/ndn/base.py
+++ b/eos_data_distribution/ndn/base.py
@@ -34,13 +34,6 @@ from pyndn import Name, Data, Face, MetaInfo
 
 logger = logging.getLogger(__name__)
 
-
-def makeName(o):
-    if isinstance(o, Name):
-        return o
-    return Name(o)
-
-
 class GLibUnixTransport(UnixTransport):
     _watch_id = 0
 
@@ -114,6 +107,7 @@ class Base(GObject.GObject):
 
         self._callbackCount = 0
         self._responseCount = 0
+
 
 class Producer(Base):
     __gsignals__ = {
--- a/eos_data_distribution/ndn/base.py
+++ b/eos_data_distribution/ndn/base.py
@@ -145,7 +145,7 @@ class Producer(Base):
 
     def sendFinish(self, data):
         # self.sign(data)
-        # logger.debug ('sending data: %d', data.__len__())
+        logger.debug('sending data: %s', data.getName())
         self.face.putData(data)
 
     def _onInterest(self, *args, **kwargs):
--- a/eos_data_distribution/ndn/base.py
+++ b/eos_data_distribution/ndn/base.py
@@ -136,7 +136,7 @@ class Producer(Base):
     def generateKeys(self):
         # Use the system default key chain and certificate name to sign commands.
         self._key_chain = KeyChain()
-        self._cert_name = keyChain.getDefaultCertificateName()
+        self._cert_name = self._key_chain.getDefaultCertificateName()
         self.face.setCommandSigningInfo(self._key_chain, self._cert_name)
 
     def sign(self, data):
--- a/eos_data_distribution/ndn/base.py
+++ b/eos_data_distribution/ndn/base.py
@@ -201,8 +201,6 @@ class Consumer(Base):
 
     def _onData(self, interest, data):
         self._callbackCount += 1
-        # that is complicated… getContent() returns an ndn.Blob, that needs
-        # to call into buf() to get a bytearray…
         self.emit('data', interest, data)
 
     def expressInterest(self, name=None, forever=False):
--- a/eos_data_distribution/ndn/chunks.py
+++ b/eos_data_distribution/ndn/chunks.py
@@ -103,6 +103,7 @@ class Consumer(base.Consumer):
 
         super(Consumer, self).__init__(name=name, *args, **kwargs)
         self.connect('data', self._on_data)
+        logger.debug('init chunks.Consumer: %s', name)
 
     def start(self):
         # Make an initial request for the barename. We should get a fully
@@ -135,6 +136,7 @@ class Consumer(base.Consumer):
 
     def _request_segment(self, n):
         ndn_name = Name(self._qualified_name).appendSegment(n)
+        logger.debug('is this an interest ? %s', ndn_name)
         self.expressInterest(ndn_name, forever=True)
         if self._segments is not None:
             self._segments[n] = SegmentState.OUTGOING
--- a/eos_data_distribution/ndn/chunks.py
+++ b/eos_data_distribution/ndn/chunks.py
@@ -96,12 +96,17 @@ class Consumer(base.Consumer):
         self._num_segments = None
         self._segments = None
         self._num_outstanding_interests = 0
+        self._qualified_name = None
 
         super(Consumer, self).__init__(name=name, *args, **kwargs)
         self.connect('data', self._on_data)
 
     def start(self):
-        self._request_segment(0)
+        # Make an initial request for the barename. We should get a fully
+        # qualified request back for the first segment, with a timestamp and
+        # segment number. Future requests will request the fully qualified
+        # name.
+        self.expressInterest(self.name, forever=True)
 
     def _save_chunk(self, n, data):
         pass
@@ -126,7 +131,7 @@ class Consumer(base.Consumer):
             self._request_segment(next_segment)
 
     def _request_segment(self, n):
-        ndn_name = Name(self.name).appendSegment(n)
+        ndn_name = Name(self._qualified_name).appendSegment(n)
         self.expressInterest(ndn_name, forever=True)
         if self._segments is not None:
             self._segments[n] = SegmentState.OUTGOING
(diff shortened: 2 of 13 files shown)


eos_data_distribution/ndn/base.py
```diff
@@ -118,9 +118,9 @@ def removeRegisteredPrefix(self, prefix):
 
     def registerPrefix(self, prefix=None, postfix="", flags=None):
         prefix = makeName(prefix)
-        postfix = makeName(postfix)
 
         if not prefix:
+            postfix = makeName(postfix)
```
magcius: You can drop this entirely -- names will get constructed automatically if the thing you pass to `.append` isn't a Name already.

eos_data_distribution/ndn/chunks.py
```diff
@@ -125,12 +125,15 @@ def _schedule_interests(self):
...
+    def _request(self, name):
+        self.expressInterest(name, forever=True)
+
     def _request_segment(self, n):
         ndn_name = Name(self.name).appendSegment(n)
-        self.expressInterest(ndn_name, forever=True)
         if self._segments is not None:
             self._segments[n] = SegmentState.OUTGOING
         self._num_outstanding_interests += 1
+        self._request(ndn_name)
```
magcius: Don't see the need for this change.

eos_data_distribution/ndn/http.py
```diff
@@ -54,17 +54,27 @@ def read_bytes_async():
...
-    content_range = msg.response_headers.get_one('Content-Range')
+    return msg.response_headers
+
+def get_content_size(headers):
+    content_range = headers.get_one('Content-Range')
     size = int(content_range.split('/')[1])
     return size
 
+def get_last_modified(headers):
+    # XXX - we should probably try to read that, but as strptime is local
```
magcius: "read that"?

eos_data_distribution/ndn/http.py
```diff
@@ -84,7 +94,9 @@ def __init__ (self, url,
 
         # XXX -- this is a bit ugly that we're making an HTTP request
         # in the constructor here...
-        self._size = get_content_size(self._session, self.url)
+        self._headers = get_headers (self._session, self.url)
```
magcius: No space between function name and (

magcius:
I'm not a big fan of the code in this branch -- it breaks several abstractions and is difficult for me to read.

eos_data_distribution/ndn/http.py
xaiki: yeah, convert to int, i'll get it now.

eos_data_distribution/ndn/chunks.py
xaiki: DRY, we need request on first packet to get the right name back. we could call expressInterest here, but i'd rather link back to the same code

eos_data_distribution/ndn/http.py
xaiki: done.

eos_data_distribution/ndn/http.py
```diff
@@ -68,12 +68,20 @@ def get_content_size(headers):
...
-    # XXX - we should probably try to read that, but as strptime is local
-    # dependent it makes it annoyingly hard to do so, anyway, this string
-    # should be good enough.
     # note that we can't use ETag as we need things to be ordered
-    return headers.get_one('Last-Modified')
+    g = lm_re.search(headers.get_one('Last-Modified')).groups()
+    d = int(g[0])
+    M = MonthMapping.index(g[1]) + 1
+    (y, h, m, s) = (int(i) for i in g[2:])
+    return "%04d%02d%02d:%02d%02d%02dGMT" % (y, M, d, h, m, s)
```
magcius: What's wrong with `strptime(lmStr, "%a, %d %b %Y %H:%M:%S GMT")`?

xaiki: doc says it's dependent on the locale…

magcius: Yikes. This code still isn't... great... but I'll take it for now.

eos_data_distribution/ndn/http.py
```diff
@@ -99,8 +100,19 @@ def _got_stream(self, msg, task, data):
...
     def _got_buf(self, data, buf):
         data.setContent(buf)
-        self.sendFinish(data)
+        self.emit('data', data)
 
+class Producer(chunks.Producer):
+    def __init__(self, name, url, session=None, *args, **kwargs):
+        super(Producer, self).__init__(name, *args, **kwargs)
+        self._getter = Getter (url, session)
+        self._getter.connect ('data', lambda o, d: self.sendFinish (d))
```
pwithnall: Nitpick: Non-PEP-8 style throughout this commit.

eos_data_distribution/ndn/test.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/python
+# -*- Mode:python; coding: utf-8; c-file-style:"gnu"; indent-tabs-mode:nil -*- */
+#
+# Copyright (C) 2016 Endless Computers INC.
```
pwithnall: I think this should be ‘Endless Mobile, Inc.’? At least, nobody’s corrected me on this yet.

eos_data_distribution/ndn/test.py
```diff
@@ -0,0 +1,65 @@
...
+import re
+import argparse
+
+import logging
+logging.basicConfig(level=logging.DEBUG)
+
+def process_args(*extra):
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-n", "--name")
+    parser.add_argument("-o", "--output")
```
pwithnall: Nitpick: Missing `help=` for these two.

eos_data_distribution/ndn/http.py
pwithnall: Nitpick: s/local/locale/

eos_data_distribution/ndn/manifest.py
```diff
@@ -0,0 +1,51 @@
...
+
+from pyndn import Name
+
+from . import http
+from . import chunks
+
+logger = logging.getLogger(__name__)
+
+# the manifest producer is an http Producer that answers on a different name
+class Producer (chunks.Producer):
```
pwithnall: Nitpick: Code isn’t PEP 8 compliant. Could you hook up `pep8` to your editor?

eos_data_distribution/ndn/http.py
pwithnall: `soup_date_new_from_string()` should do what you want.

xaiki: cool, that works.

eos_data_distribution/ndn/test.py
xaiki: done.

eos_data_distribution/ndn/test.py
xaiki: that's to small a nitpick, it's explicit enough.

eos_data_distribution/ndn/manifest.py
xaiki: done.

xaiki:
once #8 is merged this applies

magcius:
Pushed a rebased form of this which uses some of my cleanups. I'll get to reviewing it soon.

eos_data_distribution/ndn/base.py
```diff
@@ -115,6 +115,10 @@ def __init__(self, name, face=None):
         self._callbackCount = 0
         self._responseCount = 0
 
+    def setName(self, name):
+        self.name = Name(name)
```
magcius: I don't really like this -- if we're going the full "a producer owns one name", I feel like that name should be immutable. If we check the prefixes in the future, this makes it possible to do `registerPrefix(nameA)`, then `setName(nameB)`, and now we own "multiple names".

The existence of this was one reason I started ripping out the Producer class, basically, as I felt it was a cheat.

eos_data_distribution/ndn/chunks.py
magcius: `self._request` already is a fairly simple method that I don't see much value in extracting. Requesting a barename and requesting a chunk are two very different things IMO.

magcius: I think `self._request()` should request for the barename and the internals should just use `self.expressInterest`

eos_data_distribution/ndn/manifest.py
```diff
@@ -0,0 +1,51 @@
...
+# the manifest producer is an http Producer that answers on a different name
+class Producer(chunks.Producer):
+    def __init__(self, name, url, session=None, *args, **kwargs):
+        self._getter = http.Getter(url, session)
+        self._getter.connect('data', lambda o, d: self.sendFinish(d))
+
+        # XXX -- we mangle the name in the constructor, this is slow
+        self._reply_name = Name(name).append(self._getter._last_modified)
+
+        super(Producer, self).__init__(name, *args, **kwargs)
```
magcius: Why can't you pass `self._reply_name` to the parent constructor?

eos_data_distribution/ndn/manifest.py
```diff
@@ -0,0 +1,51 @@
...
+logger = logging.getLogger(__name__)
+
+# the manifest producer is an http Producer that answers on a different name
+class Producer(chunks.Producer):
+    def __init__(self, name, url, session=None, *args, **kwargs):
+        self._getter = http.Getter(url, session)
+        self._getter.connect('data', lambda o, d: self.sendFinish(d))
+
+        # XXX -- we mangle the name in the constructor, this is slow
+        self._reply_name = Name(name).append(self._getter._last_modified)
```
magcius: Any reason we shouldn't turn this feature on for everything using the HTTP producer? Seems weird to limit this to only be the manifest producer.

magcius: Also, how will this work with the store? We're going to have to have the store decode the timestamp header, compare the dates, and do the same song and dance too, right? For USB and P2P and such to work correctly, the store has to be able to check timestamp requests.

As such, please use the 0xFC marker (`name.appendTimestamp`) so we can detect that in the store.

eos_data_distribution/ndn/http.py
```diff
@@ -118,6 +120,10 @@ def _got_buf(self, data, buf):
     else:
         name = re.sub('https?://', '', args.url)
 
+    loop = GLib.MainLoop()
     producer = Producer(name, args.url, auto=True)
+    if args.output:
+        consumer = file.FileConsumer(name, filename=args.output, auto=True)
+        consumer.connect('complete', lambda  *a: loop.quit())
```
magcius: The user could simply run `edd-getchunks` now, right? Any reason for both?

eos_data_distribution/ndn/http.py
```diff
@@ -54,17 +54,25 @@ def read_bytes_async():
 
     read_bytes_async()
 
-def get_content_size(session, url):
+def get_headers(session, url):
     # XXX: SOMA's subscriptions-frontend doesn't handle HEAD requests yet because S3
     # is a bit silly with signed requests. For now, request a bytes=0-0 range and
     # look at the Content-Range.
```
magcius: Comment needs updating.

eos_data_distribution/ndn/http.py
```diff
@@ -82,7 +90,9 @@ def __init__(self, url, session=None, chunk_size=chunks.CHUNK_SIZE):
 
         # XXX -- this is a bit ugly that we're making an HTTP request
         # in the constructor here...
-        self._size = get_content_size(self._session, self.url)
+        self._headers = get_headers(self._session, self.url)
```
magcius: `get_headers` is a bit generic of a function name here, and `get` usually doesn't imply network activity. A different name would be better here...

eos_data_distribution/ndn/base.py
magcius: Also, I don't see this used anywhere?

xaiki: agreed. it should go.

eos_data_distribution/ndn/manifest.py
xaiki: because in this case we need to reply on another name (A lower name) than the one we own.

eos_data_distribution/ndn/manifest.py
xaiki: the store asks for the name without the timestamp asking for the <right most child> the higher date wins, no need to know about the date. it's not really a timestamp, it should be a version number, but we don't have that.

eos_data_distribution/ndn/http.py
xaiki: `fetch_http_headers` ?

eos_data_distribution/ndn/http.py
xaiki: it was to have an all-inclusive test, but yes, i think this is where your modules entry points make sense.

eos_data_distribution/ndn/chunks.py
xaiki: i'm not sure i get this comment, `_request()` is the internals that uses self.expressInterest ?

eos_data_distribution/ndn/base.py
xaiki: removed

eos_data_distribution/ndn/http.py
xaiki: updated

eos_data_distribution/ndn/http.py
xaiki: done.

eos_data_distribution/ndn/base.py
xaiki: done (by you)

eos_data_distribution/ndn/manifest.py
magcius: Why can't we own the lower name, though?

eos_data_distribution/ndn/manifest.py
magcius: I'm talking about the non-internet-connected case. If we don't have any edge router, then the store also needs to know how to satisfy a request for the latest timestamp with the highest timestamp in its store that matches, correct?

xaiki: Sure, whatever provides the rightmost value wins.

The store will keep the full path so will the USB producen. And we finally have real named data.
Inteligence should be provided if we get a partial file from the store but can't get it anymore from http. I guess we need to timeout the store request and re-request without the timestamp bit.

magcius: "Whatever provides the rightmost value wins"

I still don't understand. The SOMA subscriptions fetcher might ask for `/com.endlessm/subscriptions/soma/v1/1052abcd/shards/foo.shard/time=Dec-29-2016`, right? But the on-disk OSTree store only has `/com.endlessm/subscriptions/soma/v1/1052abcd/shards/foo.shard/time=Dec-25-2016`. The store has to have some logic to satisfy the former request with the latter, right?

eos_data_distribution/ndn/base.py
```diff
@@ -203,13 +203,13 @@ def _onData(self, interest, data):
...
-            name = self.name
+    def expressInterest(self, interest=None, forever=False):
+        if interest is None:
+            interest = self.name
 
-        logger.debug("Express Interest name: %s", name)
-        onTimeout = partial(self.onTimeout, forever=forever, name=name)
-        self.pit[name] = self.face.expressInterest(name, self._onData, onTimeout)
+        logger.debug("Express Interest name: %s", interest)
+        onTimeout = partial(self.onTimeout, forever=forever, name=self.name)
```
magcius: This doesn't make sense to me -- if `onTimeout` gets called, it will request another interest with the base name of the consumer, rather than the one that actually timed out. This seems wrong to me -- we should fix up `onTimeout`.

xaiki: yes that's the hack i mentioned on slack, what we need to do here is remove mustBeFresh and re-request the same interest.

eos_data_distribution/ndn/test.py
pwithnall: Really? What is being named? What format should the name follow? What’s an output? An output file? Consider that all users are going through the same ramp-up as me.

PR #6, opened 2016-12-22
T14639-debian-master

pwithnall:
This is the part of T14639 which adds Debian packaging support. It’s in a separate branch because it needs to end up as the `debian-master` branch eventually, not merged into `master`.


debian/changelog
```diff
@@ -0,0 +1,12 @@
+eos-data-distribution (0.1.0) UNRELEASED; urgency=medium
+
+  [ Philip Withnall ]
+  * Initial Debian directory and packaging
+
+  [ Jasper St. Pierre ]
+  * (many changes, please see version control system)
```
xaiki: no need for that, debian changelog is for debian/ changelogs.

debian/changelog
```diff
@@ -0,0 +1,12 @@
...
+
+  [ Niv Sardi ]
+  * (many changes, please see version control system)
```
xaiki: ditto, not trying to steal light from jasper =)

debian/control
```diff
@@ -0,0 +1,42 @@
...
+ dh-python,
+ python-all,
+ python-pyndn,
+ python-pytest,
+ python-setuptools,
+ python-setuptools-git,
+ python3-all,
+ python3-pyndn,
+ python3-setuptools,
+ python3-setuptools-git,
```
xaiki: this looks a bit weird, did you pass `wrap-and-sort` on it ?

debian/control
```diff
@@ -0,0 +1,42 @@
...
+Standards-Version: 3.9.7
+Testsuite: autopkgtest-pkg-python
+X-Python-Version: >= 2.7
+X-Python3-Version: >= 3.0
```
xaiki: add links to git repos here.

debian/control
```diff
@@ -0,0 +1,42 @@
...
+
+Package: python-eosdatadistribution
```
xaiki: did i mention i really don't like that name ?

debian/control
```diff
@@ -0,0 +1,42 @@
...
+Depends:
+ ${python:Depends},
+ ${misc:Depends},
+Description: NDN producers and daemons (Python 2)
+ Named Data Networking provides name-based access to data over the
+ network.
+ .
+ This package contains the services and library for Python 2.
+
+Package: python3-eosdatadistribution
```
xaiki: we need to split that into more packages i think, at least:
 - lib
 - dbus service
 - ostree-store
 - usb stuff

debian/copyright
```diff
@@ -0,0 +1,41 @@
+Format: http://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
+Upstream-Name: eos-data-distribution
+
+Files: *
+Copyright:
+ © 2016 Endless Mobile, Inc.
+License: LGPL-3+
+
+Files:
+ dbus-service.py
```
xaiki: that copyright is bad, it's been written by jasper...

pwithnall:
> [ ] integration for dbus activation of the dbus service
> [ ] integration for udev/udisk/whatever for the usb producer
> [ ] integration for systemd integration for the store and edge-router (now soma-subscritpion.)

These will follow in later commits. This commit is just to get what we have already packaged up.

debian/control
pwithnall: I wrapped and sorted it manually (the perils of doing all your development on a non-Debian machine). I’ll double-check it.

debian/changelog
pwithnall: OK.

debian/control
pwithnall: OK.

debian/control
pwithnall: We could fix this by splitting it out into things like `python-edd-utils`, `python-edd-stores`, `python-edd-producers`, etc. — but see my comment below.

debian/control
pwithnall: Realistically, would any of these be useful by themselves? If not, I’m not sure they should be separate packages.

debian/copyright
pwithnall: In the file? One of you two should update the copyright then (probably better than if I do it, since I’m a third party here)

debian/control
magcius: Agree. I don't see any benefit to splitting into multiple packages.

debian/copyright
magcius: Indeed. We have a lot of code written in-house that copies the UCLA declaration. In fact, I don't think there's any UCLA-owned code in our repo at all.

magcius:
Your PR here is set to be merged into "master" and has edits to the Python code. I don't know what happened but this seems wrong to me.

pwithnall:
> Your PR here is set to be merged into "master" and has edits to the Python code. I don't know what happened but this seems wrong to me.

Whoops, fixed.

debian/control
xaiki: having the edge-router in it's own package certainly has benefits, as we don't need to enable this behaviour for all the instalations and i'm not sure we can get a better mechanism in time for mid-january.

debian/copyright
xaiki: just like god, you're one of us ! *creepy* *folk* *guitar* *sounds*

pwithnall: To be clear: changing the copyrights in files should happen in a separate commit. I’m not going to do it in this commit, as it would confuse matters. The separate commit should update `debian/copyright`.

pwithnall:
Rebased against T14639 and updated. Changes include `wrap-and-sort` and `debian/changelog` trimming.

Things which need to be done as follow-up commits:
 * Updating the copyrights throughout the project (and updating `debian/copyright` accordingly)

magcius:
> pwithnall wants to merge 1 commit into T14639 from T14639-debian-master

This still seems wrong to me. It should want to merge into debian-master from T14639-debian-master

pwithnall:
> This still seems wrong to me. It should want to merge into debian-master from T14639-debian-master

There is no `debian-master` branch yet, so github won’t let me set it as the merge destination. What’s the right way of doing this? Push a `debian-master` branch for the same commit as `master`, then update github?

sgnn7:
@pwithnall `git checkout --orphan <new_branch>`

pwithnall:
Pushed `debian-master` to `origin` as a new orphan branch. Thanks for the help.

PR #7, opened 2016-12-22
Xaiki/fix get subscription

xaiki:
apply changes requested by @pwithnall

### DIFF
--- a/eos_data_distribution/ndn/base.py
+++ b/eos_data_distribution/ndn/base.py
@@ -115,7 +115,6 @@ class Base(GObject.GObject):
         self._callbackCount = 0
         self._responseCount = 0
 
-
 class Producer(Base):
     __gsignals__ = {
         'register-failed': (GObject.SIGNAL_RUN_FIRST, None, (object, )),
@@ -129,9 +128,9 @@ class Producer(Base):
         self.generateKeys()
         self._prefixes = dict()
 
-        if auto: self.produce()
+        if auto: self.start()
 
-    def produce(self):
+    def start(self):
         self.registerPrefix()
 
     def generateKeys(self):
@@ -208,9 +207,10 @@ class Consumer(Base):
         self.pit = dict()
         #        self.generateKeys()
         self._prefixes = dict()
-        if auto: self.consume()
 
-    def consume(self):
+        if auto: self.start()
+
+    def start(self):
         self.expressInterest()
 
     def _onData(self, interest, data):
--- a/eos_data_distribution/parallel.py
+++ b/eos_data_distribution/parallel.py
@@ -33,8 +33,9 @@ class Batch(GObject.GObject):
         super(Batch, self).__init__()
         self._type = type
         self._incomplete_workers = set(workers)
-        for batch in self._incomplete_workers:
-            batch.connect('complete', self._on_batch_complete)
+        for worker in self._incomplete_workers:
+            worker.connect('complete', self._on_batch_complete)
+            worker.start()
 
     def _on_batch_complete(self, batch):
         logger.info("%s complete: %s", (self._type, batch))
--- a/get_subscription.py
+++ b/get_subscription.py
@@ -1,6 +1,6 @@
 # -*- Mode:python; coding: utf-8; c-file-style:"gnu"; indent-tabs-mode:nil -*- */
 #
-# Copyright (C) 2016 Endless Computers INC.
+# Copyright (C) 2016 Endless Mobile INC.
 # Author: Niv Sardi <xaiki@endlessm.com>
 #
 # This program is free software: you can redistribute it and/or modify
--- a/get_subscription.py
+++ b/get_subscription.py
@@ -81,7 +81,7 @@ except:
     store_dir = "./eos_subscription_data"
 
 parser = argparse.ArgumentParser()
-parser.add_argument("-t", "--store-dir", default=store_dir)
+parser.add_argument("-t", "--store-dir", default=store_dir, type=str, help="where to store the downloaded files")
 parser.add_argument("appids", nargs='+')
 
 args = parser.parse_args()
--- a/get_subscription.py
+++ b/get_subscription.py
@@ -94,7 +94,7 @@ except KeyError as e:
 
 assert subids.__len__()
 
-fetchers = [Fetcher(args.store_dir, subid, face=face).start() for subid in subids]
+fetchers = [Fetcher(args.store_dir, subid, face=face) for subid in subids]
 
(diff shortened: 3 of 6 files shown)


magcius:
Please follow a consistent commit message format that contains which files or components you're touching, and why you are making the change you are.

get_subscription.py
```diff
@@ -1,3 +1,22 @@
+# -*- Mode:python; coding: utf-8; c-file-style:"gnu"; indent-tabs-mode:nil -*- */
+#
+# Copyright (C) 2016 Endless Computers INC.
```
pwithnall: Nitpick: ©

I think our legal entity is ‘Endless Mobile, Inc.’, so you should probably use that. At least, that’s what’s on the website and in the copyrights for things like `eos-desktop`.

get_subscription.py
```diff
@@ -66,13 +66,15 @@
 
 def mount_get_root(mount):
     drive = mount.get_drive()
+    if not drive or not drive.is_removable():
+        return False
```
pwithnall: I would return `None` here, since the function otherwise returns an object, and we probably want to keep the types compatible.

get_subscription.py
```diff
@@ -81,7 +81,7 @@ def mount_get_root(mount):
     store_dir = "./eos_subscription_data"
 
 parser = argparse.ArgumentParser()
-parser.add_argument("-t", "--store-dir", default=store_dir)
+parser.add_argument("-t", "--store-dir", default=store_dir, type=str, help="where to store the downloaded files")
 parser.add_argument("appids", nargs='+')
```
pwithnall: `help` is needed for `appids` too.

get_subscription.py
```diff
@@ -87,12 +87,14 @@ def mount_get_root(mount):
 args = parser.parse_args()
 
 try:
-    appids = [app_to_sub[s] for s in args.appids]
+    subids = [app_to_sub[a] for a in args.appids if re.match('^\w+$', a)]
```
pwithnall: It would be good to split this regexp out into a utility function in one of the libraries, both so that it can be reused elsewhere, and so that it has a method name which makes it a bit more obvious what it does.

Note that the regexp as it stands is equivalent to the character class `[a-zA-Z0-9_]`, which is not ‘all ASCII characters’ as Jasper suggested; but this is fine by me. We can always make it less restrictive later if needed.

get_subscription.py
xaiki: fixed

get_subscription.py
xaiki: let's split it when we need to use it. also this should be fine, ids are alphanum

xaiki:
this depends on #8

get_subscription.py
magcius: Agreed. Return `None` please.

get_subscription.py
magcius: `type=str` is the default I'm quite sure -- don't think it's needed.

get_subscription.py
magcius: ... what's the regex for? I don't understand it.

get_subscription.py
```diff
@@ -87,12 +87,14 @@ def mount_get_root(mount):
...
 except KeyError as e:
     logger.critical("couldn't find subid for app: %s", e.args)
     sys.exit()
 
-fetchers = [Fetcher(args.store_dir, appid, face=face).start() for appid in appids]
+assert subids.__len__()
+
+fetchers = [Fetcher(args.store_dir, subid, face=face).start() for subid in subids]
 
 batch = Batch(fetchers, "Subscriptions")
```
magcius: As mentioned before, there's a race here between `Batch` and `fetchers` that removing `auto=True` would fix.

get_subscription.py
magcius: I still see `Endless Computers INC.` here.

get_subscription.py
xaiki: ok, pushed 2 commits to fix that, it s/consume,produce/start/ for constency

eosdatadistribution/ndn/base.py
```diff
@@ -115,6 +115,11 @@ def __init__(self, name, face=None):
         self._callbackCount = 0
         self._responseCount = 0
 
+        if auto: self.start()
```
magcius: This is broken, and that's kind of what I meant. Since you chain up first thing in the constructor, `start()` will run before the rest of the constructor runs, leading to undefined member variable errors.

eos_data_distribution/ndn/base.py
```diff
@@ -115,6 +115,11 @@ def __init__(self, name, face=None):
         self._callbackCount = 0
         self._responseCount = 0
 
+        if auto: self.start()
```
magcius: (GitHub didn't get my comment through? Leaving it again...)

This is broken since you chain up before you construct the rest of the object. To give an example, `Consumer` chains up first, then sets `self.pit`. But since `start` here calls `expressInterest` before `self.pit` is initialized, we'll fail when we try to run `self.pit[filename] = ...`

This is why I don't like `auto=True` as a paradigm -- constructor ordering is very, very tricky to get right and I'd prefer not to deal with it at all.

eosdatadistribution/ndn/base.py
xaiki: moved it back to producer/consumer.

eos_data_distribution/ndn/base.py
xaiki: got your other comment, you are right, left start in consumer/producer, we can decide to remove it later.

eos_data_distribution/parallel.py
```diff
@@ -35,6 +35,7 @@ def __init__(self, workers, type="Batch"):
         self._incomplete_workers = set(workers)
         for batch in self._incomplete_workers:
             batch.connect('complete', self._on_batch_complete)
+            worker.start()
```
magcius: Should rename this to `for worker in self._incomplete_workers`, or this should be `batch.start()`

xaiki: right ... that name confused me...

eos_data_distribution/ndn/base.py
```diff
@@ -115,7 +115,6 @@ def __init__(self, name, face=None):
         self._callbackCount = 0
         self._responseCount = 0
 
-
```
magcius: The extra newline was intentional and is PEP8 style.

magcius:
Merged this, with a few style fixes and adjustments on top.

PR #8, opened 2016-12-23
T14639 packaging

xaiki:
we are upstream,
 - no need to break other PRs by silly name change, dh_python does the right thing...
 - no need to split debian packaging


dbus-service.py
```diff
@@ -30,8 +30,8 @@
 
 from pyndn import Name, Face
 
-from eosdatadistribution.ndn import Consumer
```
magcius: You forgot to rename the files themselves.

scripts/edd-get-appdata
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env python
```
magcius: Python convention is to have all code in packages / modules so we can test it, and then use entry_points. Not a fan of this commit.

dbus-service.py
xaiki: i don't get that comment ?

magcius: You changed the imports back to eos_data_distribution but the files are still named eosdatadistribution in this commit.

See the tree: https://github.com/endlessm/endless-ndn/tree/49266c8b01291b7feb74d051160d848554f454b7

scripts/edd-get-appdata
xaiki: i followed this: https://docs.python.org/2/distutils/setupscript.html#installing-scripts
couldn't really find a better doc, happy to read more.

magcius: See http://python-packaging.readthedocs.io/en/latest/command-line-scripts.html

dbus-service.py
xaiki: squashed both commits.

magcius:
Pushed entrypoints for the remaining scripts, except for get_subscription, since that's waiting for fixups in another PR.

PR #9, opened 2016-12-28
Xaiki/command interest

xaiki:
putting this up for review,

i'm not sure it's the right way to do this (especially don't really like command/cost to be split out) but i like to be able to test them this way... maybe @pwithnall can help with a better way to organize things while retaining testability.


eos_data_distribution/ndn/base.py
```diff
@@ -220,7 +220,7 @@ def _onData(self, interest, data):
         self.emit('data', interest, data)
 
     def makeInterest(self, name):
-        return Name(name)
+        return Interest(name)
```
magcius: We were removing this, no?

eos_data_distribution/ndn/command.py
```diff
@@ -0,0 +1,101 @@
...
+from pyndn.control_parameters import ControlParameters
+from pyndn.encoding.tlv_wire_format import TlvWireFormat
+from pyndn.util.command_interest_generator import CommandInterestGenerator
+from pyndn import Interest, Name
+
+logger = logging.getLogger(__name__)
+
+_commandInterestGenerator = CommandInterestGenerator()
+
+def makeCommandInterest (*args, **kwargs):
```
magcius: Needs PEP8 fixes.

eos_data_distribution/ndn/command.py
```diff
@@ -0,0 +1,101 @@
...
+}
+
+def makeInterest(cmd, flags=None, local=True,
+                 keyChain=None, certificateName=None,
+                 controlParameters={}):
+    cp = ControlParameters()
+
+    logger.debug('control Parameters: %s', controlParameters)
+    for c in controlParameters:
+        getattr(cp,controlMap[c])(controlParameters[c])
```
magcius: Sorry but I really don't like this kind of generic coding. Just have the creator pass in a `ControlParameters` instance.

eos_data_distribution/ndn/cost.py
```diff
@@ -0,0 +1,75 @@
...
+# A copy of the GNU Lesser General Public License is in the file COPYING.
+
+import logging
+
+from . import base
+from pyndn.node import Node
+
+logger = logging.getLogger(__name__)
+
+class Producer(base.Producer):
```
magcius: Why is this a Producer? It doesn't actually produce anything, right?

eos_data_distribution/ndn/command.py
magcius: All of this still holds.

eos_data_distribution/MDNS/__init__.py
```diff
@@ -42,8 +42,8 @@ def __init__(self):
         self.pretty_name = avahi.ServiceTypeDatabase.ServiceTypeDatabase()
 
     def get_human_type(self, type):
-        if type in self.pretty_name:
-            return self.pretty_name[type]
+        if str(type) in self.pretty_name:
+            return getattr(self.pretty_name, type)
```
magcius: This seems very wrong to me. You can't `getattr` anything but a `str`, so the `str(type)` doesn't mean anything.

eos_data_distribution/ndn/command.py
```diff
@@ -47,6 +47,17 @@ def makeCommandInterest (*args, **kwargs):
     'flagbit': 'setFlagBit'
 }
 
+def addNextHop(faceUri, name, cost=0, *args, **kwargs):
+    face = Face(faceUri)
```
magcius: This needs to be a GLibUnixFace in order for it to be integrated with the mainloop, right? I don't quite understand why we're connecting to somebody else's machine here...

edge-router-avahi.py
```diff
@@ -0,0 +1,79 @@
...
+#from gi.repository import Gio
+
+from eos_data_distribution.MDNS import ServiceDiscovery
+
+logging.basicConfig(level=logging.INFO)
+
+SERVICES = [
+    # Disable TCP, we really only want UDP or ethernet
+    # "_nfd._tcp",
+    "_nfd._udp"]
```
magcius: Why don't we want TCP?

edge-router-avahi.py
```diff
@@ -0,0 +1,79 @@
...
+
+gatways, ips = None, None
+
+
+def flatten(l):
+    return [i for s in l for i in s]
```
magcius: Unused.

edge-router-avahi.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+class EdgeRouter(object):
+    def __init__(self, *args, **kwargs):
+        super(EdgeRouter, self).__init__()
```
magcius: `super` and `*args, **kwargs` not needed here.

eos_data_distribution/ndn/chunks.py
```diff
@@ -76,6 +76,29 @@ def _on_interest(self, o, prefix, interest, face, interestFilterId, filter):
 
         self._send_chunk(data, seg)
 
+    def _registerPrefix(self, prefix, cost=None, controlParameters={},
+                           onInterest=None, onRegisterFailed=None,
+                           onRegisterSuccess=None,
+                           *args, **kwargs):
+        if not cost: cost = self.cost
```
magcius: I feel like this should be on `base`.

eos_data_distribution/MDNS/__init__.py
xaiki: didn't work without it...

edge-router-avahi.py
xaiki: not tested, but i'd say udp will work better for us... because, tcp is evil ?

eos_data_distribution/ndn/chunks.py
xaiki: so it would be this and the register interest stuff ? yep, makes sense, will move them now.

edge-router-avahi.py
xaiki: fxd

edge-router-avahi.py
xaiki: fxd

edge-router-avahi.py
magcius: I mean, the perfect fit would be SCTP but who uses that in 2017

eos_data_distribution/ndn/command.py
xaiki: oh yeah, i missed it, let me see how to solve this, i don't really like to build a control parameter elsewhere, but i guess we could

eos_data_distribution/ndn/base.py
```diff
@@ -209,15 +211,28 @@ def registerPrefix(self, prefix=None, flags=None, *args, **kwargs):
         self._prefixes[prefix] = self._registerPrefix(prefix, flags, *args, **kwargs)
         return prefix
 
-    def _registerPrefix(self, prefix, flags=None,
+    def _registerPrefix(self, prefix, cost=None, controlParameters={},
```
magcius: This will cause bugs. Don't use mutable values as default parameters, as you only get one object. Use `=None` and construct it inside.

eos_data_distribution/ndn/command.py
xaiki: fxd, it actually looks much nicer you were right.

edge-router-avahi.py
xaiki: for our case it'd be better to use the ethernet transport, but i don't think you're going to like the libpcap dependency ;)

eos_data_distribution/ndn/base.py
```diff
@@ -211,17 +213,17 @@ def registerPrefix(self, prefix=None, flags=None, *args, **kwargs):
...
                            *args, **kwargs):
         if not cost: cost = self.cost
         if not onInterest: onInterest = self._onInterest
         if not onRegisterFailed: onRegisterFailed = self.onRegisterFailed
         if not onRegisterSuccess: onRegisterSuccess = self.onRegisterSuccess
+        if not controlParameters: controlParameters = ControlParameters()
 
-        if cost: controlParameters['cost'] = int(cost)
-        logger.debug("Register Prefix: %s", controlParameters)
+        if cost: controlParameters.setCost(int(cost))
```
magcius: Do you ever pass a non-integer cost here? I don't see the need for the cast.

eos_data_distribution/ndn/base.py
xaiki: fixed by subsequent commits.

eos_data_distribution/ndn/base.py
xaiki: when it comes from the command line, i could cast it earlier, but here i'm sure i'm getting them all (if no cast i need an assert.

eos_data_distribution/MDNS/__init__.py
pwithnall: > didn't work without it...

Do you know why?

edge-router-avahi.py
pwithnall: > not tested, but i'd say udp will work better for us... because, tcp is evil ?

What’s the reasoning here? It would be useful to have the reasoning in the comment in the code here.

eos_data_distribution/ndn/base.py
pwithnall: If it comes from the command line you should do the cast at the same time you check its validity; and you should probably reject any inputs which are not valid integers.

eos_data_distribution/ndn/command.py
```diff
@@ -0,0 +1,90 @@
+# -*- Mode:python; coding: utf-8; c-file-style:"gnu"; indent-tabs-mode:nil -*- */
+#
+# Copyright (C) 2016 Endless Mobile INC.
```
pwithnall: Nitpick: Copyright holder

eos_data_distribution/ndn/command.py
```diff
@@ -0,0 +1,90 @@
...
+
+    commandInterest = Interest()
+    if local:
+        commandInterest.setName(Name("/localhost%s" % cmd))
+        # The interest is answered by the local host, so set a short timeout.
+        commandInterest.setInterestLifetimeMilliseconds(2000.0)
+    else:
+        commandInterest.setName(Name("/localhop%s" % cmd))
+        # The host is remote, so set a longer timeout.
+        commandInterest.setInteresLifetimeMilliseconds(4000.0)
```
pwithnall: s/setInteresLifetimeMilliseconds/setInterestLifetimeMilliseconds/

eos_data_distribution/ndn/command.py
```diff
@@ -0,0 +1,90 @@
...
+
+def makeInterest(cmd, local=True, controlParameters={},
+                 keyChain=None, certificateName=None):
+    assert(cmd.startswith('/'))
+
+    commandInterest = Interest()
+    if local:
+        commandInterest.setName(Name("/localhost%s" % cmd))
+        # The interest is answered by the local host, so set a short timeout.
+        commandInterest.setInterestLifetimeMilliseconds(2000.0)
```
pwithnall: How did you choose this magic number (and the one below)?

eos_data_distribution/ndn/command.py
```diff
@@ -0,0 +1,90 @@
...
+        commandInterest.setInteresLifetimeMilliseconds(4000.0)
+    # NFD only accepts TlvWireFormat packets.
+    commandInterest.getName().append(controlParameters.wireEncode(TlvWireFormat.get()))
+    generateInterest(commandInterest,
+                     keyChain, certificateName,
+                     TlvWireFormat.get())
+
+    return commandInterest
+
+if __name__ == '__main__':
```
pwithnall: Move all of the code in here into a `main()` method which you can then call from unit tests to check the command line parsing.

eos_data_distribution/ndn/command.py
```diff
@@ -0,0 +1,90 @@
...
+    from gi.repository import GLib
+    from . import base
+    import argparse
+
+    logging.basicConfig(level=logging.DEBUG)
+
+    parser = argparse.ArgumentParser(description='Command Interest Tests')
```
pwithnall: Why not use `unittest` for the tests, then they can be run automatically when setuptools are run and when the package is built?

eos_data_distribution/ndn/cost.py
```diff
@@ -0,0 +1,49 @@
+# -*- Mode:python; coding: utf-8; c-file-style:"gnu"; indent-tabs-mode:nil -*- */
+#
+# Copyright (C) 2016 Endless Mobile INC.
```
pwithnall: Nitpick: Copyright holder

eos_data_distribution/ndn/cost.py
```diff
@@ -0,0 +1,49 @@
...
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+# A copy of the GNU Lesser General Public License is in the file COPYING.
+
+import logging
+
+from . import base
+
+logger = logging.getLogger(__name__)
+
+if __name__ == '__main__':
```
pwithnall: As above, all the code here should be moved into a `main()` method, or moved into a `unittest` file (ideally the latter — automated tests are much more valuable than manual ones).

eos_data_distribution/ndn/cost.py
```diff
@@ -0,0 +1,49 @@
...
+    logging.basicConfig(level=logging.DEBUG)
+
+    parser = argparse.ArgumentParser(description='Register with cost Test')
+    parser.add_argument("name")
+    parser.add_argument("-c", "--cost", default=10)
+    args = parser.parse_args()
+
+    name = args.name
+    if not name:
+        name = Name('/endless/test')
```
pwithnall: Doesn’t this mean that after this branch, the type of `name` will either be `str` or `Name`? Can `Producer()` and `registerPrefix()` deal with that?

eos_data_distribution/ndn/file.py
```diff
@@ -250,3 +255,16 @@ def _on_complete(self):
...
         super(FileConsumer, self)._on_complete()
+
+if __name__ == '__main__':
+    from . import test
+    args = process_args("filename")
+
+    if args.name:
+        name = args.name
+    else:
+        name = args.filename
```
pwithnall: Could be simplified to `name = args.name or args.filename`, I think?

eos_data_distribution/ndn/http.py
```diff
@@ -53,42 +55,52 @@ def read_bytes_async():
...
+    return msg.response_headers
+
+def get_content_size(headers):
+    content_range = headers.get_one('Content-Range')
     size = int(content_range.split('/')[1])
     return size
 
+def get_last_modified(headers):
+    # note that we can't use ETag as we need things to be ordered
+    date = Soup.Date.new_from_string(headers.get_one('Last-Modified'))
```
pwithnall: What if there is no `Last-Modified` header? `headers.get_one()` will return `None` and `Date.new_from_string()` will probably be sad.

eos_data_distribution/ndn/http.py
```diff
@@ -99,25 +111,31 @@ def _got_stream(self, msg, task, data):
 
     def _got_buf(self, data, buf):
         data.setContent(buf)
-        self.sendFinish(data)
+        self.emit('data', data)
```
pwithnall: (I haven’t profiled but…) I suspect emitting a signal for every chunk of data received is going to add considerable CPU overhead. GObject signal emissions are not cheap.

eos_data_distribution/ndn/http.py
```diff
@@ -99,25 +111,31 @@ def _got_stream(self, msg, task, data):
...
+        logger.debug ('getter: got buf')
 
+class Producer(chunks.Producer):
+    def __init__(self, name, url, session=None, *args, **kwargs):
+        super(Producer, self).__init__(name, cost=DEFAULT_COST, *args, **kwargs)
+        self._getter = Getter(url, session=session, chunk_size=self.chunk_size)
```
pwithnall: Should `assert(chunk_size > 0)` somewhere (possibly in the parent constructor?), because we divide by the chunk size in `_get_final_segment()`.

eos_data_distribution/ndn/http.py
```diff
@@ -99,25 +111,31 @@ def _got_stream(self, msg, task, data):
...
+    def _get_final_segment(self):
+        return self._getter._size // self.chunk_size
+
+    def _send_chunk(self, data, n):
+        self._getter.soup_get(data, n)
 
 if __name__ == '__main__':
-    from gi.repository import GLib
     import re
+    from . import test
```
pwithnall: I think these would be better as `unittest`s, so they can be picked up by setuptools and autopkgtest.

eos_data_distribution/ndn/manifest.py
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/python
+# -*- Mode:python; coding: utf-8; c-file-style:"gnu"; indent-tabs-mode:nil -*- */
+#
+# Copyright (C) 2016 Endless Computers INC.
```
pwithnall: Nitpick: Copyright holder

eos_data_distribution/ndn/test.py
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/python
+# -*- Mode:python; coding: utf-8; c-file-style:"gnu"; indent-tabs-mode:nil -*- */
+#
+# Copyright (C) 2016 Endless Computers INC.
```
pwithnall: Nitpick: Copyright holder

eos_data_distribution/ndn/test.py
```diff
@@ -0,0 +1,50 @@
...
+    else:
+        logging.basicConfig(level=logging.DEBUG)
+
+    return args
+
+def run_test(args, name):
+    loop = GLib.MainLoop()
+
+    if args.output:
+        consumer = file.FileConsumer(name, filename=args.output, auto=True)
```
pwithnall: To automate these tests (`unittest`), you should be able to hardcode `name` and rework `FileConsumer` a bit so you can pass a `file`-like object (from `tempfile.NamedTemporaryFile`) to it instead of `filename`. That will allow the unit tests to use temporary files while avoiding race conditions with temporary filename generation.

eos_data_distribution/ndn/test.py
```diff
@@ -0,0 +1,50 @@
...
+
+    args = parser.parse_args()
+    if args.v == 0:
+        logging.basicConfig(level=logging.INFO)
+    else:
+        logging.basicConfig(level=logging.DEBUG)
+
+    return args
+
+def run_test(args, name):
```
pwithnall: Instead of utility functions, this lot could be a `unittest.TestCase` subclass.

eos_data_distribution/soma_subscription_fetcher.py
```diff
@@ -57,12 +57,18 @@ def __init__(self):
...
     def _on_interest(self, o, prefix, interest, face, interestFilterId, filter):
-        # We handle one fo two paths:
+        # We handle one of two paths:
         #   /com.endlessm/subscriptions/soma/subscription/$sub_id/manifest.json/$chunk
         #   /com.endlessm/subscriptions/soma/shard/$shard_url/$chunk
 
         name = interest.getName()
-        chunkless_name = name.getPrefix(-1)
+
+        # XXX: Use more sophisticated parsing algorithm to strip non-chunk parts.
```
pwithnall: You might as well split this code out into such a helper function now (even if the algorithm remains this simple), so that it can be easily reused in future without propagating more `XXX`s throughout the code.

eos_data_distribution/tools/util.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/python
+# -*- Mode:python; coding: utf-8; c-file-style:"gnu"; indent-tabs-mode:nil -*- */
+#
+# Copyright (C) 2016 Endless Computers INC.
```
pwithnall: Nitpick: Copyright holder

eos_data_distribution/ndn/base.py
```diff
@@ -114,6 +111,61 @@ def __init__(self, name, face=None):
...
+        self.pit[interest] = self.face.expressInterest(interest, onData, onTimeout)
+        return interest
+
+    def makeCommandInterest(self, cmd, prefix=None, controlParameters=None,
+                            keyChain=None, certificateName=None,
+                            *args, **kwargs):
+        if not prefix: prefix=self.name
+        if not controlParameters: controlParameters = ControlParameters()
+        if not self._keyChain or not self._certificateName:
+            self.generateKeys()
```
pwithnall: Do we need to call `self.generateKeys()` if we are not going to use `self._keyChain` and `self._certificateName` because the caller has provided `keyChain` and `certificateName`?

eos_data_distribution/tools/util.py
xaiki: fxd

eos_data_distribution/soma_subscription_fetcher.py
xaiki: fxd

eos_data_distribution/ndn/test.py
xaiki: fxd

eos_data_distribution/ndn/manifest.py
xaiki: fxd

eos_data_distribution/ndn/http.py
xaiki: fxd

eos_data_distribution/ndn/http.py
xaiki: ok, fixed

eos_data_distribution/ndn/http.py
xaiki: fxd

eos_data_distribution/ndn/file.py
xaiki: fxd

eos_data_distribution/ndn/cost.py
xaiki: it should, names were always quite interchangeably `str` or `Name`, they even became `Interest` without much notice, we're fixing this last part.

eos_data_distribution/ndn/cost.py
xaiki: fxd

eos_data_distribution/ndn/cost.py
xaiki: fxd

eos_data_distribution/ndn/command.py
xaiki: fxd

eos_data_distribution/ndn/command.py
xaiki: commented, it comes from PyNDN2

eos_data_distribution/ndn/command.py
xaiki: fxd

eos_data_distribution/ndn/base.py
```diff
@@ -194,6 +245,16 @@ def onRegisterSuccess(self, prefix, registered):
...
         logger.info("Register succeded for prefix: %s, %s", prefix, registered)
 
+    def removeRegisteredPrefix(self, prefix):
+        name = Name(prefix)
+        logger.info("Un-Register prefix: %s", name)
+        try:
+            self.face.removeRegisteredPrefix(self._prefixes[name])
+            del (self._prefixes[name])
+        except:
+            logger.warning("tried to unregister a prefix that never was registred: %s", prefix)
```
xaiki: fxd

eos_data_distribution/ndn/base.py
```diff
@@ -194,6 +245,16 @@ def onRegisterSuccess(self, prefix, registered):
         self.emit('register-success', prefix, registered)
         logger.info("Register succeded for prefix: %s, %s", prefix, registered)
 
+    def removeRegisteredPrefix(self, prefix):
+        name = Name(prefix)
+        logger.info("Un-Register prefix: %s", name)
+        try:
+            self.face.removeRegisteredPrefix(self._prefixes[name])
+            del (self._prefixes[name])
```
xaiki: fxd

eos_data_distribution/MDNS/__init__.py
```diff
@@ -95,21 +98,16 @@ def service_resolved(self, interface, protocol, name, type, domain, host, aproto
...
+              (name, type, flags, domain, self.siocgifname(interface), avahi.LOOKUP_RESULT_LOCAL))
 
         # this check is for local services
-        #        try:
-        #            if flags & avahi.LOOKUP_RESULT_LOCAL:
-        #                return
-        #        except dbus.DBusException:
-        #            pass
+        if flags & avahi.LOOKUP_RESULT_LOCAL:
+            logger.info ('droping local service')
```
xaiki: fxd

edge-router-avahi.py
```diff
@@ -0,0 +1,75 @@
...
+        ifname = sda.siocgifname(interface)
+
+    def network_changed_cb(monitor, available):
+        print "network changed"
+        gatways = netifaces.gateways()
+
+
+if __name__ == "__main__":
+    #    nm = Gio.NetworkMonitor.get_default()
+    #    nm.connect('network-changed', network_changed_cb)
```
xaiki: not used anymore, this was an old tool that did something quite different (it was the edge-router now named eos_subscription_producer)

edge-router-avahi.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/python
+# -*- Mode:python; coding: utf-8; c-file-style:"gnu"; indent-tabs-mode:nil -*- */
+#
+# Copyright (C) 2016 Endless Computers INC.
```
xaiki: fxd

xaiki:
@pwithnall implemented your suggestions minus the ones touching edge-router-avahi (that is still very much WIP) and unitest (that belong in a separate PR)

eos_data_distribution/ndn/http.py
```diff
@@ -111,14 +109,13 @@ def _got_stream(self, msg, task, data):
...
-        logger.debug ('getter: got buf')
+        self.onData(data);
 
 class Producer(chunks.Producer):
     def __init__(self, name, url, session=None, *args, **kwargs):
         super(Producer, self).__init__(name, cost=defaults.RouteCost.HTTP, *args, **kwargs)
-        self._getter = Getter(url, session=session, chunk_size=self.chunk_size)
-        self._getter.connect('data', lambda o, d: self.sendFinish(d))
+        self._getter = Getter(url, session=session, chunk_size=self.chunk_size,
+                              onData=lambda d: self.sendFinish(d))
```
pwithnall: This could just be `onData=self.sendFinish`, right?

eos_data_distribution/ndn/chunks.py
```diff
@@ -34,9 +34,19 @@ def get_chunk_component(name):
     # The chunk component of a name is the last part...
     return name.get(-1)
 
+def get_chunkless_name(name):
```
pwithnall: I haven’t really said it anywhere else yet, but in general it would be nice to see more pydoc comments around,  especially on functions like this which do something fairly impenetrable. They’re easy enough to write quickly when putting the function together. :-)

eos_data_distribution/ndn/command.py
```diff
@@ -70,22 +70,39 @@ def makeInterest(cmd, local=True, controlParameters={},
...
     logging.basicConfig(level=logging.DEBUG)
 
-    parser = argparse.ArgumentParser(description='Command Interest Tests')
-    parser.add_argument("-n", "--name")
+    parser = utils.process_args(description='Command Interest Tests')
+    parser.add_argument("--faceid", "-i")
+    parser.add_argument("--uri", "-u")
+    parser.add_argument("--local-control-feature", "-l")
+    parser.add_argument("--origin", "-o")
+    parser.add_argument("--cost", "-c")
```
pwithnall: `type=int` here will do the type checking for you. Similarly for the other arguments.

eos_data_distribution/ndn/command.py
```diff
@@ -70,22 +70,39 @@ def makeInterest(cmd, local=True, controlParameters={},
...
     from gi.repository import GLib
     from . import base
-    import argparse
+    from tests import utils
 
     logging.basicConfig(level=logging.DEBUG)
 
-    parser = argparse.ArgumentParser(description='Command Interest Tests')
-    parser.add_argument("-n", "--name")
+    parser = utils.process_args(description='Command Interest Tests')
```
pwithnall: General comment about this commit (“ndn: prepare commands to be testable”): I have no idea what is going on here. What are your plans for the testing, and why aren’t they discussed in the commit message? What do all these arguments do, and why don’t they have `help` to explain it?

pwithnall:
> @pwithnall implemented your suggestions minus the ones touching edge-router-avahi (that is still very much WIP) and unitest (that belong in a separate PR)

@xaiki: that sounds reasonable, thanks. I have a few more suggestions above, and would be interested to see your plans for the unit tests. :-)

eos_data_distribution/MDNS/__init__.py
xaiki: it complained getattr only supports strings IIRC.

eos_data_distribution/ndn/command.py
xaiki: fixd

eos_data_distribution/ndn/manifest.py
```diff
@@ -33,11 +33,13 @@
 # the manifest producer is an http Producer that answers on a different name
 class Producer(chunks.Producer):
     def __init__(self, name, url, session=None, *args, **kwargs):
-        self._getter = http.Getter(url, session)
-        self._getter.connect('data', lambda o, d: self._send_finish(d))
+        self._getter = http.Getter(url, onData=self._send_finish, session=session)
+        self._last_modified = http.get_last_modified(self._getter._headers)
+        if self._last_modified == None:
```
pwithnall: Nitpick: I think `if not self._last_modified:` is more pythonic, but this works equally well.

avahi-monitor.py
```diff
@@ -36,6 +38,16 @@
     # "_nfd._tcp",
     "_nfd._udp"]
 
+def face_uri_from_triplet(type, host, port):
+    if type == '_nfd._udp':
+        proto = 'udp4'
+    else:
+        proto = 'tcp4'
+    return "%s://%s:%s"%(proto, host, port)
```
pwithnall: Nitpick: Is the `//` necessary here? Technically, the scheme and address of a URI only have to be separated by a colon (`:`).

pwithnall: Nitpick: PEP 8 requires spaces around the `%`. Can you set up your editor to highlight PEP 8 violations?

avahi-monitor.py
```diff
@@ -36,6 +38,16 @@
...
+
+def build_registery_key(name, type, domain):
+    return "%s-%s-%s"%(name, type, domain)
```
pwithnall: Nitpick: PEP 8 requires spaces around the `%`. Can you set up your editor to highlight PEP 8 violations?

avahi-monitor.py
```diff
@@ -36,6 +38,16 @@
...
     "_nfd._udp"]
 
+def face_uri_from_triplet(type, host, port):
+    if type == '_nfd._udp':
+        proto = 'udp4'
+    else:
+        proto = 'tcp4'
+    return "%s://%s:%s"%(proto, host, port)
+
+def build_registery_key(name, type, domain):
```
pwithnall: Nitpick: s/registery/registry/

avahi-monitor.py
```diff
@@ -47,16 +59,23 @@ def __init__(self):
 
         sda.start()
         self.sda = sda
+        self._registery = dict()
```
pwithnall: Nitpick: s/registery/registry/

avahi-monitor.py
```diff
@@ -47,16 +59,23 @@ def __init__(self):
...
 
     def service_added_cb(self, sda, interface, protocol, name, type, h_type, domain, host, aprotocol, address, port, txt, flags):
         ifname = sda.siocgifname(interface)
         print "Found Service data for service '%s' of type '%s' (%s) in domain '%s' on %s.%i:" % (name, h_type, type, domain, ifname, protocol)
-        command.addNextHop(faceURI, cost=defaults.RouteCost.LOCAL_NETWORK)
+        faceUri = face_uri_from_triplet(type, host, port)
+        check_call(["nfdc", "add-nexthop",
+                    "-c", "%d"%defaults.RouteCost.LOCAL_NETWORK,
```
pwithnall: Nitpick: `str(defaults.RouteCost.LOCAL_NETWORK)` would be a bit simpler.

avahi-monitor.py
```diff
@@ -47,16 +59,23 @@ def __init__(self):
...
+                    "%s"%SUBSCRIPTIONS_SOMA,
```
pwithnall: Nitpick: Similarly, `str(SUBSCRIPTIONS_SOMA)` would be a bit simpler.

avahi-monitor.py
```diff
@@ -47,16 +59,23 @@ def __init__(self):
...
         sda.start()
         self.sda = sda
+        self._registery = dict()
 
     def service_added_cb(self, sda, interface, protocol, name, type, h_type, domain, host, aprotocol, address, port, txt, flags):
         ifname = sda.siocgifname(interface)
         print "Found Service data for service '%s' of type '%s' (%s) in domain '%s' on %s.%i:" % (name, h_type, type, domain, ifname, protocol)
-        command.addNextHop(faceURI, cost=defaults.RouteCost.LOCAL_NETWORK)
+        faceUri = face_uri_from_triplet(type, host, port)
+        check_call(["nfdc", "add-nexthop",
```
pwithnall: Do you want to handle `CalledProcessError` from this, or are you deliberately letting it propagate up to the caller?

avahi-monitor.py
```diff
@@ -47,16 +59,23 @@ def __init__(self):
...
+                    "%s"%SUBSCRIPTIONS_SOMA,
+                    faceUri])
+        self._registery[build_registery_key(name, type, domain)] = faceUri
 
     def service_removed_cb(self, sda, interface, protocol, name, type, domain, flags):
         ifname = sda.siocgifname(interface)
         print "Disappeared Service '%s' of type '%s' in domain '%s' on %s.%i." % (name, type, domain, ifname, protocol)
-        command.removeNextHop(faceURI, cost=defaults.RouteCost.LOCAL_NETWORK)
+        faceUri = self._registery[build_registery_key(name, type, domain)]
+        check_call(["nfdc", "remove-nexthop", "%s"%SUBSCRIPTIONS_SOMA, faceUri])
```
pwithnall: Same comments apply here.

eos_data_distribution/names.py
```diff
@@ -19,5 +19,5 @@
 
 from pyndn import Name
 
-SUBSCRIPTIONS_INSTALLED = Name('/com.endlessm/subscriptions/installed')
-SUBSCRIPTIONS_SOMA = Name('/com.endlessm/subscriptions/soma')
+SUBSCRIPTIONS_INSTALLED = Name('/endlessm/subscriptions/installed')
```
pwithnall: I think it would be worthwhile to put some of the explanation from your commit message into a comment here, so it’s easy to find.

eos_data_distribution/ndn/command.py
```diff
@@ -143,7 +133,12 @@ def makeInterest(cmd, local=True, controlParameters={},
         commandInterest.setName(Name("/localhop%s" % cmd))
         # The host is remote, so set a longer timeout.
         commandInterest.setInterestLifetimeMilliseconds(4000.0)
+    if interestLifeTime: commandInterest.setInterestLifetimeMilliseconds(interestLifeTime)
     # NFD only accepts TlvWireFormat packets.
+
+    logger.debug("DUMP: %s", dumpControlParameter(controlParameters))
+    logger.debug("wire encoding: %s", str(controlParameters.wireEncode(TlvWireFormat.get())).encode('quopri'))
```
pwithnall: Why did you choose quoted-printable encoding here? UTF-8 would be more compact and readable in the printouts, and is supported by all consoles.

avahi-monitor.py
xaiki: added autopep8...

xaiki: it is, this is translated into a management command in the faceUri field, and it *hasto* be complete. https://redmine.named-data.net/projects/nfd/wiki/FaceMgmt#FaceUri

avahi-monitor.py
xaiki: done.

avahi-monitor.py
xaiki: done

avahi-monitor.py
xaiki: something very bad happens if we can't add next-hop, might as well die

avahi-monitor.py
xaiki: ok

avahi-monitor.py
xaiki: ok

avahi-monitor.py
xaiki: ok

eos_data_distribution/ndn/command.py
xaiki: because we have special chars, utf8 just show: case case box x

eos_data_distribution/ndn/base.py
xaiki: it is removed iirc.

eos_data_distribution/ndn/command.py
xaiki: we have 315 pep8 violations according to pep8, can we merge and then make a styling commit ?

eos_data_distribution/ndn/cost.py
xaiki: removed

eos_data_distribution/ndn/command.py
xaiki: we're not, we're connecting via a GLibUnixFace to NFD telling it to connect to another machine,
has been implemented differently and removed.

edge-router-avahi.py
xaiki: tcp is slow and unpredictable ?

eos_data_distribution/ndn/base.py
xaiki: fixed

xaiki:
all comments fixed, can we have a review @magcius (and merge) plez ?

pwithnall:
> we have 315 pep8 violations according to pep8, can we merge and then make a styling commit ?

Seems reasonable. I don’t know where github has hidden the original comment where you said this though.

xaiki:
@pwithnall last round of review and merge ?

pwithnall:
> @pwithnall last round of review and merge ?

Looks OK to me, but as I’ve repeatedly said, I’m not qualified enough with the code base to actually ack a merge.

PR #10, opened 2017-01-05
T13934

pwithnall:
The first round of changes needed for [T13934](https://phabricator.endlessm.com/T13934), to add a systemd `.service` file and udev integration for `edd-usb-producer` so that it runs when a USB stick is plugged in.

It will now run when a new block device appears, and will automatically quit after 30s if there are no interesting mounts around (and if no content is currently published from any mounts).

I have done some basic testing with a USB stick containing an empty `.endless-NDN-DATA` directory to check the system integration; but I have not yet put together a full `.endless-NDN-DATA` directory and tested it that way. I will do that and submit any resulting fixes as a separate PR, as they are less to do with the system integration.

### DIFF
--- a/debian/python-eos-data-distribution.install
+++ b/debian/python-eos-data-distribution.install
@@ -1,0 +1,3 @@
+eos_data_distribution/producers/70-edd-usb-producer.rules usr/lib/udev/rules.d
+eos_data_distribution/producers/edd-usb-producer.service lib/systemd/system
+
--- a/eos_data_distribution/producers/usb.py
+++ b/eos_data_distribution/producers/usb.py
@@ -42,10 +42,8 @@ IDLE_TIMEOUT = 30  # seconds
 
 
 def is_mount_interesting(mount):
-    """Is the given mount interesting: does it have an associated drive,
-       and does it contain NDN cache data?"""
-    return (mount.get_drive() and
-            path.exists(path.join(mount.get_root().get_path(),
+    """Is the given mount interesting: does it contain NDN cache data?"""
+    return (path.exists(path.join(mount.get_root().get_path(),
                                   ENDLESS_NDN_CACHE_PATH)))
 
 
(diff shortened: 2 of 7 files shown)


debian/python-eos-data-distribution.install
```diff
@@ -0,0 +1,3 @@
+eos_data_distribution/producers/70-edd-usb-producer.rules usr/lib/udev/rules.d
+eos_data_distribution/producers/edd-usb-producer.service lib/systemd/system
```
xaiki: i'm pretty sure we should split this out into many packages.

eos_data_distribution/producers/usb.py
```diff
@@ -35,36 +37,84 @@
...
+                mount.get_name(), mount.get_uuid() or "no UUID")
+    store.publish_all_names(base)
 
 
 def mount_removed_cb(monitor, mount, store):
     root = mount.get_root()
-    p = root.get_path()
-    pprint.pprint([store.remove_name(n) for p, n in store.producers.items() if p.startswith(p)])
+    root_path = root.get_path()
+    pprint.pprint([store.remove_name(n) for p, n in store.dirpubs.items() if p.startswith(root_path)])
```
xaiki: that pprint is old code, it should probably go away before we ship it.

eos_data_distribution/producers/usb.py
```diff
@@ -35,36 +37,84 @@
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
 
+
+# Timeout after which this daemon will exit if no interesting mounts exist.
+IDLE_TIMEOUT = 30  # seconds
```
xaiki: shouldn't this go into defaults.py ?

xaiki:
fixups should be squashed

debian/python-eos-data-distribution.install
pwithnall: As discussed on the call today, it’s best to do that later (if at all), if the existing monolithic package starts pulling in too many dependencies across the system; i.e. only if there’s a good reason to want to install part of eos-data-distribution without any of the rest of it. At the moment, I think it’s reasonable to install it all together.

eos_data_distribution/producers/usb.py
pwithnall: I’ll do a follow-up commit shortly.

eos_data_distribution/producers/usb.py
pwithnall: Only if it’s going to be used by more than one module, which at the moment is not the case. We can always move it across later, but for the moment I think it makes sense for it to stay in the file where it’s used.

pwithnall:
> fixups should be squashed

oh bother, I meant to do that before creating the PR. :-(

debian/python-eos-data-distribution.install
xaiki: i thought we talked about splitting the source into many packages. but ok, let's not be blocked by that.

eos_data_distribution/producers/usb.py
xaiki: ok

PR #11, opened 2017-01-13
Xaiki/package all the things

xaiki:
First run at making a debian package for our NDN babies.


debian/control
```diff
@@ -27,4 +27,4 @@ Description: NDN producers and daemons (Python 2)
  Named Data Networking provides name-based access to data over the
  network.
  .
- This package contains the services and library for Python 2.
+ This package contains the services and library for Python 2.
```
pwithnall: Actually, this change is entirely reasonable. Ignore me.

PR #15, opened 2017-01-16
PEP8 ALL THE THINGS

xaiki:
**DO NOT MERGE BEFORE THE DEMO**

*please base all future NDN work on this branch*

(this was done using autopep8)

Signed-off-by: Niv Sardi <xaiki@evilgiggle.com>

### DIFF
--- a/eos_data_distribution/ndn/file.py
+++ b/eos_data_distribution/ndn/file.py
@@ -14,6 +14,7 @@ def get_file_size(f):
 
 
 class FileProducer(chunks.Producer):
+
     def __init__(self, name, file, *args, **kwargs):
         super(FileProducer, self).__init__(name, *args, **kwargs)
         self.name = name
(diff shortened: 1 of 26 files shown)


eos_data_distribution/ndn/file.py
```diff
@@ -82,25 +83,32 @@ def mkdir_p(dirname):
 SEGMENT_TABLE_MAGIC = 'EosSgtV1'
 
 # 7 => [0, 0, 0, 0, 0, 1, 1, 1]
+
+
```
pwithnall: I think a better fix here would be to put the comment text in a docstring for the function, rather than effectively moving it further away from the function it documents.

eos_data_distribution/ndn/file.py
```diff
@@ -82,25 +83,32 @@ def mkdir_p(dirname):
...
 def num_to_bitmap(n):
     assert 0 <= n <= 255
     return [int(c) for c in list(bin(n)[2:].zfill(8))]
 
 # [0, 0, 0, 0, 0, 1, 1, 1] -> 7
+
+
```
pwithnall: Similarly: I think a better fix here would be to put the comment text in a docstring for the function, rather than effectively moving it further away from the function it documents.

xaiki:
it was all done with autopep8

pwithnall:
This was merged into `next` as PR#20, without fixing the review comments. :-(

xaiki:
ooops, missed those, have them stacked ontop of xaiki/performance

PR #16, opened 2017-01-16
T14639

pwithnall:
* Fix logging output in `edd-dbus-service`
 * Fix a crash in `edd-usb-producer` due to a base directory argument not being specified


xaiki:
please rebase on pep8 and ask for merge there so that nobody merges that into master by accident before the demo.

eos_data_distribution/SimpleStore/__init__.py
```diff
@@ -40,9 +41,9 @@ def __init__(self, base=None, prefix='/',
 
         if base: self.publish_all_names(base)
 
-    def _path_to_name(self, filename):
-        assert filename.startswith(self.base)
-        file_path = filename[len(self.base):].lstrip('/')
+    def _path_to_name(self, filename, basedir):
+        assert filename.startswith(basedir)
```
xaiki: maybe we should do realpath() here ? now that method has a subtle bug if called in another contexct than init -> publish_all_names()

eos_data_distribution/producers/usb.py
```diff
@@ -107,7 +107,6 @@ def main():
     GLib.unix_signal_add(GLib.PRIORITY_DEFAULT, signal.SIGTERM, signal_cb)
     monitor = Gio.VolumeMonitor.get()
     store = SimpleStoreProducer(prefix=SUBSCRIPTIONS_SOMA,
-                                split=defaults.ENDLESS_NDN_CACHE_PATH,
```
xaiki: you need split !!!
this is what maps `/media/blah/usb/some/dir/soma/ndn/files` to -> `/soma/ndn/files`
this looks very wrong.

eos_data_distribution/SimpleStore/__init__.py
pwithnall: I’m not so sure. If the base path contains a symlink somewhere, all the paths produced by `walk()` will do as well (I believe), so I don’t think we want to call `realpath()` in `_path_to_name()` or the prefix assertion will fail.

eos_data_distribution/producers/usb.py
pwithnall: `self.split` was never used in `SimpleStore.Producer`. I think the mapping in `usb.py` was implemented by the fact that `publish_all_names()` is called with a different base path each time.

eos_data_distribution/SimpleStore/__init__.py
xaiki: @magcius put the realpath because it broke producing on the right name, we should wait for his input on it.

eos_data_distribution/producers/usb.py
xaiki: so you are going to need to do at least that for usb (or you'll break it) we can use base instead of split and deprecate split.

pwithnall: I don’t understand how this change can break anything: the value of `split` was never used anywhere.

xaiki: that means that probably other changes broke usb...

xaiki: but by the looks of it , usb is broken..

magcius:
The demo isn't focused on USB. I probably broke it to get other stuff to work. I didn't understand the need for both `split` and `base`, so I made everything just use `base` in there. We can do the same for the USBProducer too.

eos_data_distribution/producers/usb.py
magcius: The demo isn't focused on USB. I probably broke it to get other stuff to work. I didn't understand the need for both split and base, so I made everything just use base in there. We can do the same for the USBProducer too.

pwithnall: I suspect these cleanup commits need to be dropped and re-done by someone who knows what on earth is going on in this code. I can guess (no documentation or comments! :( ), but apparently I will guess wrong.

xaiki: i cherry-picked your dbus_service fix in my performance branch, (that you validated) there is another fix for base crash too. maybe we can drop that branch and merge the other ?

pwithnall: That’s PR #17; I’ve left some more review comments there. Let’s leave this PR #16 open until that’s resolved.

xaiki:
i think this is all merged in #17 can we close @pwithnall  ?

pwithnall:
Yes.

PR #17, opened 2017-01-16
Xaiki/performance

xaiki:
On Going fixes coming from performance work.


eos_data_distribution/subscription.py
```diff
@@ -99,13 +99,15 @@ def _on_shards_complete(self, parallel_consumer):
...
-# downloads the manifest and shards by fetching from /endless/soma/v1/foo/...,
-# and then generates a "signalling response" for them.
-
 
 class Producer(object):
 
+    """
+    # The Producer listens for intents to /endless/installed/foo,
+    # downloads the manifest and shards by fetching from /endless/soma/v1/foo/...,
+    # and then generates a "signalling response" for them.
```
pwithnall: Lines in docstring comments don’t start with `#`.

eos_data_distribution/DirTools/__init__.py
```diff
@@ -20,6 +20,11 @@ def __init__(self, dir, flags=Gio.FileMonitorFlags.NONE, userdata=None):
                 'Asked to monitor something that is not a dir: %s' % (dir, ))
         self.monitorAll(f)
 
+    # XXX: this is needed because older version of Gio.Monitor implement
+    # different signals.
+    def filterSignals(self, a):
+        return [e for e in a if e in EventToSignal.keys()]
```
pwithnall: Something like the following might be clearer, although you should check the types:
```
return set(a).intersection(EventToSignal.keys())
```

eos_data_distribution/subscription.py
xaiki: they look weird without them... but fixing..

eos_data_distribution/DirTools/__init__.py
xaiki: types ? i.e. assert that all items in a are str ?

pwithnall: i.e. You might need to convert between `list`, `set` and `iterable` (last one is Python 3 only, iirc) to get things to behave.

eos_data_distribution/ndn/base.py
```diff
@@ -139,10 +139,12 @@ def generateKeys(self, name=None):
     def sign(self, data):
         return self._keyChain.sign(data, self._certificateName)
 
-    def expressInterest(self, interest=None, *args, **kwargs):
+    def expressInterest(self, interest=None, name=None, *args, **kwargs):
         if interest is None:
             interest = Interest(self.name)
```
pwithnall: Shouldn’t this be changed to use `interest = Interest(name)` after `name` has been set up?

eos_data_distribution/DirTools/__init__.py
pwithnall: Merging fixups into the original commit would also make git archaeology a lot easier.

eos_data_distribution/SimpleStore/__init__.py
```diff
@@ -27,7 +27,7 @@
 
 class Producer(object):
 
-    def __init__(self, base=None, prefix='/',
+    def __init__(self, base='.', prefix='/',
```
pwithnall: I’m not sure we should be encouraging use of relative paths here — where do we set the `cwd`? I think it would be more explicit to require that `base` be passed in. The caller can always pass in `'.'` if it’s set up the `cwd` appropriately.

eos_data_distribution/subscription.py
pwithnall: I’ve been following http://www.sphinx-doc.org/en/stable/ext/example_google.html.

demo/mock-usb-producer.py
```diff
@@ -24,7 +24,7 @@
 from gi.repository import GLib
 
 from eos_data_distribution.names import SUBSCRIPTIONS_SOMA
-from eos_data_distribution.SimpleStore import Producer as SimpleStoreProducer
+from eos_data_distribution.store import simple_store
```
pwithnall: The commit message for this change could do with explaining why this migration is happening.

eos_data_distribution/ndn/base.py
xaiki: it should be `name or self.name`

xaiki: looking at the code it looks like `name` should actually go away...

xaiki: e4cfba6

demo/mock-usb-producer.py
xaiki: i thought having code in __init__ was not pep8 compilant ?

pwithnall: Sounds like a good reason to put in a commit message.

eos_data_distribution/subscription.py
```diff
@@ -83,7 +83,8 @@ def _fetch_manifest_complete(self, consumer):
                 path.join(self._store_dir, 'shard', escaped_filename))
             self._shard_entries.append(
                 {'manifest_path': shard['path'], 'cache_path': shard_filename})
-            consumer = FileConsumer(shard_ndn_name, shard_filename, face=self._face)
+            consumer = FileConsumer(
+                shard_ndn_name, shard_filename, face=self._face)
```
pwithnall: The changes in this commit are just formatting changes; they don’t match the commit message at all. Bad rebase/merge?

eos_data_distribution/subscription.py
```diff
@@ -105,9 +105,9 @@ def _on_shards_complete(self, parallel_consumer):
 class Producer(object):
 
     """
-    # The Producer listens for intents to /endless/installed/foo,
-    # downloads the manifest and shards by fetching from /endless/soma/v1/foo/...,
-    # and then generates a "signalling response" for them.
+    The Producer listens for intents to /endless/installed/foo,
+    downloads the manifest and shards by fetching from /endless/soma/v1/foo/...,
+    and then generates a "signalling response" for them.
```
pwithnall: This should be squashed into the original commit, 31c19c5d.

eos_data_distribution/DirTools/__init__.py
```diff
@@ -23,7 +23,7 @@ def __init__(self, dir, flags=Gio.FileMonitorFlags.NONE, userdata=None):
     # XXX: this is needed because older version of Gio.Monitor implement
     # different signals.
     def filterSignals(self, a):
-        return [e for e in a if e in EventToSignal.keys()]
+        return set(a).intersection(EventToSignal.keys())
```
pwithnall: This should be squashed into the original commit, f44fbfd5.

eos_data_distribution/SimpleStore/__init__.py
pwithnall: Commit c1405228 basically supersedes this one entirely, so this one can be dropped.

eos_data_distribution/ndn/file.py
```diff
@@ -274,6 +264,44 @@ def _on_complete(self):
         os.unlink(self._sgt_filename)
         super(FileConsumer, self)._on_complete()
```
pwithnall: `super(Consumer, self)._on_complete()` here?

eos_data_distribution/ndn/file.py
```diff
@@ -274,6 +264,44 @@ def _on_complete(self):
...
+        self._part_fd = os.open(
+            self._part_filename, os.O_CREAT | os.O_WRONLY | os.O_NONBLOCK, 0o600)
+        self._sgt_filename = '%s.sgt' % (filename, )
+        self._sgt_fd = os.open(
+            self._sgt_filename, os.O_CREAT | os.O_RDWR, 0o600)
+
+
+class FileConsumer(Consumer):
+
+    """Write to a File"""
```
pwithnall: I don’t think that’s suitable as documentation. Either remove it (so sphinx says that `FileConsumer` needs documenting), or expand it. At the moment it’s too short to be useful, but will make `FileConsumer` disappear from the lists of things which need documenting.

eos_data_distribution/ndn/file.py
```diff
@@ -274,6 +264,44 @@ def _on_complete(self):
...
+
+    def __init__(self, name, filename, *args, **kwargs):
+        self._create_files(filename)
+
+        super(FileConsumer, self).__init__(name, *args, **kwargs)
+
+
+class DirConsumer(Consumer):
+
+    """Write to a Directory routing the name"""
```
pwithnall: Similarly, this docstring should be expanded or removed.

eos_data_distribution/ndn/file.py
```diff
@@ -274,6 +264,44 @@ def _on_complete(self):
...
 
+    def _create_files(self, filename):
+        self._filename = filename
+        mkdir_p(os.path.dirname(filename))
+        self._part_filename = '%s.part' % (filename, )
+        self._part_fd = os.open(
+            self._part_filename, os.O_CREAT | os.O_WRONLY | os.O_NONBLOCK, 0o600)
+        self._sgt_filename = '%s.sgt' % (filename, )
+        self._sgt_fd = os.open(
+            self._sgt_filename, os.O_CREAT | os.O_RDWR, 0o600)
```
pwithnall: Technically, if one of the I/O operations which is not the first one raises an exception, the directory will be left in an unknown state (some of the files could exist; others won’t). It would be good to clean that up so the whole function is effectively atomic wrt I/O. This could be done as a follow-up commit, as the code here is simply moved from elsewhere.

eos_data_distribution/ndn/file.py
```diff
@@ -274,6 +264,44 @@ def _on_complete(self):
...
+        self._dirname = dirname
+        super(DirConsumer, self).__init__(name, *args, **kwargs)
+
+    def _on_data(o, interest, data):
+        self._create_files(
+            path.join(self._dirname, chunks.get_chunkless_name(
+                interest.getName())))
+
+        super(DirConsumer, self)._on_data(o, interest, data)
+        self._on_data = super(DirConsumer, self)._on_data
```
pwithnall: This seems like a hacky way to approach creating the files the first time data is received. How about always trying to create them, and ignoring the case where they already exist?

eos_data_distribution/ndn/file.py
```diff
@@ -274,6 +264,44 @@ def _on_complete(self):
         os.unlink(self._sgt_filename)
         super(FileConsumer, self)._on_complete()
 
+    def _create_files(self, filename):
+        self._filename = filename
+        mkdir_p(os.path.dirname(filename))
```
pwithnall: This is where the dirfd stuff I was talking about could be added: when creating the leaf directory, keep an FD to it, and then use the Python equivalent of `openat()` to open the files beneath it. This removes race conditions in resolving the directory path to an inode when creating each of the files. (Note that I haven’t actually thought about the code past this musing.)

eos_data_distribution/ndn/file.py
xaiki: then we'll have ghost files.

xaiki: then we get io ops on each packet !? that's a great waist of resources.

xaiki: _on_data() is basically our hotest path, this is done so that we don't do anything we don't need there.

eos_data_distribution/subscription.py
xaiki: fixed.

eos_data_distribution/DirTools/__init__.py
xaiki: squashed

eos_data_distribution/ndn/file.py
xaiki: fixed both

eos_data_distribution/ndn/file.py
```diff
@@ -104,9 +104,6 @@ class Consumer(chunks.Consumer):
     def __init__(self, *args, **kwargs):
         super(Consumer, self).__init__(*args, **kwargs)
 
-    # XXX: This is disgusting hackery. We need to remove auto=True.
-
-    def start(self):
```
magcius: Scheduling the interests should still remain in start(), just move the code that reads the segment table.

xaiki: moved that to chunks so we don't need mangle the constructor's start.

xaiki: done

eos_data_distribution/ndn/http.py
```diff
@@ -57,14 +57,13 @@ def read_bytes_async():
     read_bytes_async()
 
 
-def fetch_http_headers(session, url):
+def fetch_http_headers(session, url, callback):
```
pwithnall: Why are you trying to make header loading asynchronous? Why is this a hack? How does it differ from a full solution, and why is a hack justified at the moment? (I’m not arguing that a hack *isn’t* justified; I’d just like to see some explanation in the commit message.)

There’s not very much for me to go on with the review here.

One thought I have when looking at this: why does `Getter` exist? It’s a wrapper around libsoup which doesn’t add any additional functionality other than handling of the `Range` and `Content-Length` headers, which would be easy to do with helper functions. It’s only used from `http.Producer` — why not use libsoup directly there, and eliminate one level of Python callbacks on the hot path?

eos_data_distribution/ndn/http.py
```diff
@@ -85,24 +84,34 @@ def get_last_modified(headers):
...
+            self._session, self.url, self.got_http_headers)
+        logger.debug('getter init: %s', url)
+
+    def got_http_headers(self, o, msg, data=None):
+        logger.debug('got HTTP HEADERS: %s', msg.response_headers)
+        self._headers = msg.response_headers
+
         self._size = get_content_size(self._headers)
         if self._size == -1:
             raise ValueError("Could not determine Content-Size for %s" % url)
```
pwithnall: There’s no way for this `ValueError` to be caught. It should really be passed to `onHeaders` as a failure value, I think? Not entirely sure of the Pythonic way of doing this.

eos_data_distribution/ndn/manifest.py
```diff
@@ -32,19 +32,27 @@
...
         self._getter = http.Getter(
-            url, onData=self._send_finish, session=session)
-        self._last_modified = http.get_last_modified(self._getter._headers)
-        if not self._last_modified:
-            raise ValueError("Could not get Last-Modified")
-
-        # XXX -- we mangle the name in the constructor, this is slow
-        self._qualified_name = Name(name).append(self._last_modified)
+            url, session=session,
+            onData=self._send_finish, onHeaders=lambda headers: self._start(headers))
```
pwithnall: I think you could replace `lambda headers: self._start(headers)` with just `self._start`.

eos_data_distribution/soma_subscription_fetcher.py
```diff
@@ -86,11 +86,9 @@ def _on_interest(self, o, prefix, interest, face, interestFilterId, filter):
             assert filename == 'manifest.json'
             self._subproducers[key] = manifest.Producer(
                 chunkless_name, "%s/v1/%s/manifest.json" % (get_soma_server(), subscription_id), face=face)
-            self._subproducers[key].start()
```
pwithnall: I’d keep these `start()` calls around. They’re not doing any harm (an `info` message each), and it’s likely we’ll forget to re-add them when the `start()` implementations are changed to do something again.

demo/mock-usb-producer.py
xaiki: done

eos_data_distribution/ndn/file.py
xaiki: we don't, this is a consumer, we should not recieve different Data (can't recieve interests) packets with different names.
that is because once we get the first interest with a full name (i.e. timestamped) *we* ask for new interests on that name.

xaiki: to be clearer, we *need* to have 1 consumer per producer.

eos_data_distribution/store/simple_store.py
```diff
@@ -42,8 +42,7 @@ def __init__(self, base, prefix='/',
         self.dirs = dict()
         self.dirpubs = defaultdict(lambda: {})
 
-        if base:
-            self.publish_all_names(base)
+        self.publish_all_names(self.base)
```
xaiki: it's because a former commit made base mandatory, so that if base doesn't make sense.
also, yes we need to have the same paths (realpath()) or we get confused.

eos_data_distribution/ndn/base.py
```diff
@@ -142,6 +142,7 @@ def _expressInterest(self, interest, forever=False,
         if not onTimeout:
             onTimeout = partial(self.onTimeout, forever=forever)
 
+        interest.setInterestLifetimeMilliseconds(20000)
```
xaiki: all these are demo commits, i'm going to delete them. sorry you took time to review them.

eos_data_distribution/ndn/http.py
xaiki: we thought we were dropping packets because of time spent in soup, this commit was to test that theory, it needs to be repurposed.

we use it from http and manifest producers, it was a way to wrap those.

xaiki: although it's a hack because it breaks start() and makes it silently autostart. (described in the commit message)

eos_data_distribution/ndn/manifest.py
xaiki: true. fixed

eos_data_distribution/ndn/file.py
```diff
@@ -272,7 +252,42 @@ def _on_complete(self):
...
+
+class DirConsumer(Consumer):
+
+    def __init__(self, name, dirname, *args, **kwargs):
+        mkdir_p(dirname)
+        self._dirname = dirname
+        super(DirConsumer, self).__init__(name, *args, **kwargs)
+
+    def _on_data(o, interest, data):
+        self._create_files(
```
xaiki: ok

eos_data_distribution/ndn/file.py
```diff
@@ -272,7 +252,42 @@ def _on_complete(self):
...
+
+
+class DirConsumer(Consumer):
+
+    def __init__(self, name, dirname, *args, **kwargs):
+        mkdir_p(dirname)
+        self._dirname = dirname
+        super(DirConsumer, self).__init__(name, *args, **kwargs)
+
+    def _on_data(o, interest, data):
```
xaiki: done

eos_data_distribution/ndn/file.py
```diff
@@ -272,7 +252,42 @@ def _on_complete(self):
...
+        self._create_files(
+            path.join(self._dirname, chunks.get_chunkless_name(
```
xaiki: done

eos_data_distribution/soma_subscription_fetcher.py
xaiki: removed the hack code

eos_data_distribution/ndn/http.py
xaiki: removed from this branch

eos_data_distribution/ndn/http.py
xaiki: removed from this branch

eos_data_distribution/ndn/base.py
xaiki: removed from this branch

xaiki:
Gotta Catch'em All !

eos_data_distribution/ndn/file.py
pwithnall: (I’ll do this in my branch.)

pwithnall:
All caught. Merged. Thanks!

PR #19, opened 2017-01-18
T15105 unit tests

pwithnall:
Various changes to start adding unit tests, and fix the problems I’ve found while adding them. A lot of the fixes are based on very little knowledge of what the code was supposed to do, so are potentially wrong — but in that case they highlight something which should be documented better.

Some of the earlier commits here have already been informally reviewed when this branch was pushed (without a PR) earlier in the week. They have been updated for the comments left before.


pwithnall:
I suggest that we start accepting PRs to merge into `xaiki/pep8` (if they’re OK to accept), and rename `xaiki/pep8` to `master-next` or something like that, otherwise we’re going to end up with a horrible tangle of branches to pick apart and merge next week.

debian/rules
```diff
@@ -1,9 +1,23 @@
...
+ifneq "$(shell dh_listpackages | grep -- -doc)" ""
+	dh_sphinxdoc
+endif
+
+override_dh_auto_build: export http_proxy=127.0.0.1:9
+override_dh_auto_build: export https_proxy=127.0.0.1:9
+override_dh_auto_build:
+	dh_auto_build
+	PYTHONPATH=. sphinx-build -N -bhtml docs/ build/html # HTML generator
+	PYTHONPATH=. sphinx-build -N -bman docs/ build/man # Manpage generator
```
xaiki: all this is needed on top of dh_sphinxdoc ?!?! crazee.

eos_data_distribution/ndn/chunks.py
```diff
@@ -48,6 +48,27 @@ def get_chunkless_name(name):
...
+    final chunk may be smaller ([1, chunk size] bytes). Chunks are addressed
+    using the final component of the name a client has expressed an interest
+    in. If no chunk is addressed in the final component of the name, `Producer`
+    will return the first chunk of its content.
+
+    See the documentation for ``eos_data_distribution.ndn.base.Producer`` for
+    details on how interests are handled.
+    """
+
+    # FIXME: Should we rename ‘chunks’ to ‘segments’ for consistency?
```
xaiki: segments is an NDN concept, our chunks are a superclass of segments.

pwithnall: How do they differ?

PR #21, opened 2017-01-18
Hackily reimplement 524b8adc44a

wgreenberg:
This is a bit gross but it sets our _getter prior to init using the
chunk_size that'd be used anyway.


eos_data_distribution/ndn/http.py
```diff
@@ -115,9 +115,9 @@ def _got_buf(self, data, buf):
 
 class Producer(chunks.Producer):
     def __init__(self, name, url, session=None, *args, **kwargs):
-        super(Producer, self).__init__(name, cost=defaults.RouteCost.HTTP, *args, **kwargs)
-        self._getter = Getter(url, session=session, chunk_size=self.chunk_size,
+        self._getter = Getter(url, session=session, chunk_size=chunks.CHUNK_SIZE,
```
xaiki: you get the same results by not passing chunk_size as getter will do the same.

PR #22, opened 2017-01-20
T15105 unit tests more

pwithnall:
First batch of unit tests and fixes for the `Consumer` stack (`file.Consumer`, `chunks.Consumer`, etc.).

More tests to be added (see the list at the bottom of `test_file.py`), but this is a good start to get merged.

https://phabricator.endlessm.com/T15105

### DIFF
--- a/eos_data_distribution/ndn/base.py
+++ b/eos_data_distribution/ndn/base.py
@@ -76,6 +76,17 @@ class GLibUnixFace(Face):
         self._commandKeyChain = None
         self._commandCertificateName = Name()
 
+    @property
+    def usesGLibMainContext(self):
+        """
+        Indicator that this Face implementation uses a GLib main context.
+
+        This means it can be safely used in situations where the main context
+        is the poll implementation, and progress would not be made otherwise
+        without calling ``Face.processEvents()``.
+        """
+        return True
+
 
 def singleton(f):
     instance = [None]
@@ -113,7 +124,7 @@ class Base(GObject.GObject):
         # GLibUnixFace is the only Face implementation to do things in a GLib
         # main loop, so we require it.
         self.face = face or get_default_face()
-        assert isinstance(self.face, GLibUnixFace)
+        assert getattr(self.face, 'usesGLibMainContext', False)
 
         self._callbackCount = 0
         self._responseCount = 0
--- a/eos_data_distribution/ndn/base.py
+++ b/eos_data_distribution/ndn/base.py
@@ -146,12 +146,12 @@ class Base(GObject.GObject):
             interest = Interest(self.name)
         return self._expressInterest(interest, *args, **kwargs)
 
-    def _expressInterest(self, interest, forever=False,
+    def _expressInterest(self, interest, try_again=False,
                          onData=None, onTimeout=None):
         if not onData:
             onData = self._onData
         if not onTimeout:
-            onTimeout = partial(self.onTimeout, forever=forever)
+            onTimeout = partial(self.onTimeout, try_again=try_again)
 
         logger.debug("Express Interest name: %s", interest)
         self.pit[interest] = self.face.expressInterest(
@@ -338,11 +338,11 @@ class Consumer(Base):
         self.face.removePendingInterest(self.pit[name])
         del self.pit[name]
 
-    def onTimeout(self, interest, forever=False):
+    def onTimeout(self, interest, try_again=False):
         name = interest.getName()
         self._callbackCount += 1
-        self.emit('interest-timeout', interest, forever)
+        self.emit('interest-timeout', interest, try_again)
         logger.debug("Time out for interest: %s", name)
-        if forever:
+        if try_again:
             logger.info("Re-requesting Interest: %s", name)
-            self._expressInterest(interest, forever=forever)
+            self._expressInterest(interest, try_again=try_again)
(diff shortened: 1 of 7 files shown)


eos_data_distribution/ndn/base.py
```diff
@@ -330,7 +330,7 @@ def removePendingInterest(self, name):
     def onTimeout(self, interest, forever=False):
         name = interest.getName()
         self._callbackCount += 1
-        self.emit('interest-timeout', interest)
+        self.emit('interest-timeout', interest, forever)
```
xaiki: try_again is a better name than forever, but either fix it in the code or in your commit message.

eos_data_distribution/ndn/tests/test_file.py
```diff
@@ -137,6 +142,316 @@ def test_bitmap_to_num_invalid(self):
...
+
+    @property
+    def usesGLibMainContext(self):
+        return True
+
+    def setCommandSigningInfo(self, key_chain, certificate_name):
+        raise NotImplementedError('Not implemented in MockFace yet')
+
+    def expressInterest(self, interest, on_data, on_timeout):
+        self._interests[str(interest.name)] = (interest, on_data, on_timeout)
```
xaiki: the 'right way' to do this is interest.getName()

eos_data_distribution/ndn/chunks.py
```diff
@@ -219,6 +219,14 @@ def _on_data(self, o, interest, data):
             self._qualified_name = name.getPrefix(-1)
 
         seg = get_chunk_component(name).toSegment()
+
+        # Have we somehow already got this segment?
+        if self._segments[seg] == SegmentState.COMPLETE:
+            logger.debug('Ignoring data ‘%s’ as it’s already been received',
+                         name)
+            self._check_for_complete()
+            return
```
xaiki: the code is good, but note that this is potentially very wrong, we should **never** get 2 segments (from the same host, but in general too) or we're being messy with our requests, or the other part is being messy with it's nacks.

eos_data_distribution/ndn/chunks.py
```diff
@@ -159,7 +160,9 @@ def _on_complete(self):
 
     def _check_for_complete(self):
         if self._segments.count(SegmentState.COMPLETE) == len(self._segments):
-            self._on_complete()
+            if not self._emitted_complete:
+                self._emitted_complete = True
+                self._on_complete()
```
xaiki: it looks like we should log the case where this happens.

eos_data_distribution/ndn/file.py
```diff
@@ -122,7 +122,14 @@ def __init__(self, *args, **kwargs):
         except ValueError as e:
             pass
 
+    def _open_files(self):
+        raise NotImplementedError()
+
     def _save_chunk(self, n, data):
+        if self._part_fd < 0:
+            self._open_files()
```
xaiki: _save_chunk is as hot as _on_data

eos_data_distribution/ndn/file.py
```diff
@@ -264,6 +271,7 @@ def _on_complete(self):
 
     def _create_files(self, filename):
         # XXX this is racy
+        assert self._filename
```
xaiki: shouldn't that be filename ?

eos_data_distribution/ndn/base.py
pwithnall: Whoops. Fixed the commit message, and will add a follow-up patch to rename `forever` to `try_again` everywhere.

eos_data_distribution/ndn/tests/test_file.py
pwithnall: Why is that better than accessing it as a property (which is a bit less typing)? The documentation makes them look equivalent: http://named-data.net/doc/0.4.0/PyNDN2/pyndn.html#pyndn.interest.Interest

eos_data_distribution/ndn/chunks.py
pwithnall: This is a bit of defensive programming. The alternative is to write it as `assert self._segments[seg] != SegmentState.COMPLETE`, if you’re confident enough that NFD/PyNDN will always guarantee this for us. I would suggest the approach in the patch though.

eos_data_distribution/ndn/chunks.py
pwithnall: OK.

eos_data_distribution/ndn/file.py
pwithnall: I think an integer comparison should be OK to do on that path, though?

eos_data_distribution/ndn/file.py
pwithnall: Whoops, yes. I also meant to remove the `self._filename` assignment from there (it fits better in the callers).

eos_data_distribution/ndn/tests/test_file.py
xaiki: we tend to follow CCL, but i guess it's ok. http://named-data.net/doc/ndn-ccl-api/

eos_data_distribution/ndn/chunks.py
xaiki: let's be defensive

eos_data_distribution/ndn/file.py
xaiki: should be, if it's ever a problem it'll show in traces anyway.

eos_data_distribution/ndn/tests/test_file.py
pwithnall: OK, I’ll change it.

PR #23, opened 2017-01-20
Xaiki/rework avahi monitor

xaiki:
This ports the lessons learned from the honduras demo.
we only interconnect all nodes if we don't find an NDN node routing us

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -12,6 +12,7 @@ Build-Depends:
  python-avahi,
  python-dbus,
  python-docutils,
+ python-netifaces,
  python-pyndn,
  python-pytest,
  python-setuptools,
@@ -27,15 +28,15 @@ Vcs-Browser: https://github.com/endlessm/endless-ndn
 Package: python-eos-data-distribution
 Architecture: all
 Depends:
- ${misc:Depends},
- ${python:Depends},
  adduser,
- nfd,
  gir1.2-notify-0.7,
  gir1.2-soup-2.4,
+ nfd,
  python-avahi,
  python-dbus,
  python-pyndn,
+ ${misc:Depends},
+ ${python:Depends},
 Description: NDN producers and daemons (Python 2)
  Named Data Networking provides name-based access to data over the
  network.
--- a/eos_data_distribution/avahi_monitor.py
+++ b/eos_data_distribution/avahi_monitor.py
@@ -112,6 +112,8 @@ class AvahiMonitor(object):
                 self._registry.values().map(self.remove_nexthop)
                 # ..and only add one to the routers
                 self._ndn_gateways.keys().map(self.add_nexthop)
+                # ..and stop edge
+                self.stop_edge()
             else:
                 # no change: i am routed, i was routed before: no links
                 pass
@@ -122,6 +124,8 @@ class AvahiMonitor(object):
                 # need to add back old links...
                 # this is limited to MAX_PEERS in add_nexthop
                 self._registry.values().map(self.add_nexthop)
+                # ..and start edge
+                self.start_edge()
             else:
                 # nochange: wasn't routed, not routed now
                 pass
@@ -187,6 +191,12 @@ class AvahiMonitor(object):
     def remove_nexthop(self, faceUri):
         return check_call(["nfdc", "remove-nexthop",
                            str(SUBSCRIPTIONS_SOMA), faceUri])
+
+    def start_edge(self):
+        return check_call(["systemctl", "start", "edd-soma-subscriptions-producer"])
+
+    def stop_edge(self):
+        return check_call(["systemctl", "stop", "edd-soma-subscriptions-producer"])
 
     def get_gateways(self):
         gateways = netifaces.gateways()
--- a/eos_data_distribution/avahi_monitor.py
+++ b/eos_data_distribution/avahi_monitor.py
@@ -123,6 +123,16 @@ class AvahiMonitor(object):
             self._ndn_gateways.values()).intersection(self.gateways)
 
     def process_route_changes(self, prev_routed):
+        """ adds/removes hops for route state changes:
+        - !routed -> routed  : remove all peers but gateways
+        - routed  -> !routed : add MAX_PEERS peers, remove gateways
+        - !routed -> !routed : check we still have enough peers (MIN_PEERS)
+        - routed  -> routed  : do nothing (handled in the rest of the code)
+
+        :param prev_routed: previous routed state
+        :returns: None
+        :rtype: NoneType
+        """
         if self._routed:
             logger.info(
                 "Being routed by an NDN node, NOT acting as an edge router")
--- a/eos_data_distribution/avahi_monitor.py
+++ b/eos_data_distribution/avahi_monitor.py
@@ -138,8 +138,8 @@ class AvahiMonitor(object):
 
     def service_added_cb(self, sda, interface, protocol, name, type, h_type, domain, host, aprotocol, address, port, txt, flags):
         logger.debug(
-            "Found Service data for service '%s' of type '%s' (%s) in domain '%s' on %s.%i:",
-                     name, h_type, type, domain, ifname, protocol)
+            "Found Service data for service '%s' of type '%s' (%s) in domain '%s'",
+                     name, h_type, type, domain)
 
         self.refresh_network()
 
@@ -165,8 +165,8 @@ class AvahiMonitor(object):
 
     def service_removed_cb(self, sda, interface, protocol, name, type, domain, flags):
         logger.debug(
-            "Disappeared Service '%s' of type '%s' in domain '%s' on %s.%i.",
-            name, type, domain, ifname, protocol)
+            "Disappeared Service '%s' of type '%s' in domain '%s'",
+            name, type, domain)
 
         self.refresh_network()
 
--- a/eos_data_distribution/avahi_monitor.py
+++ b/eos_data_distribution/avahi_monitor.py
@@ -179,12 +179,8 @@ class AvahiMonitor(object):
         prev_routed = self._routed
         self.update_routed_status()
 
-        if not self._routed and not prev_routed:
-            # not routed, wasn't routed, just remove this link
-            self.remove_nexthop(faceUri)
-        else:
-            # all the rest handled by generic code
-            self.process_route_changes()
+        self.remove_nexthop(faceUri)
+        self.process_route_changes()
 
     def add_nexthop(self, faceUri):
         if len(self._peers) > MAX_PEERS:
--- a/eos_data_distribution/avahi_monitor.py
+++ b/eos_data_distribution/avahi_monitor.py
@@ -231,8 +231,10 @@ class AvahiMonitor(object):
 
     def refresh_network(self, *args, **kwargs):
         self.gateways = self.get_gateways()
+        flat_ifaces = [flatten(netifaces.ifaddresses(i).values())
+                       for i in netifaces.interfaces()]
         self.ips = flatten(
-            map(lambda a: [x['addr'] for x in a], [flatten(netifaces.ifaddresses(i).values()) for i in netifaces.interfaces()]))
+            map(lambda a: [x['addr'] for x in a], flat_ifaces))
 
     def check_call(self, a):
         try:
--- a/eos_data_distribution/avahi_monitor.py
+++ b/eos_data_distribution/avahi_monitor.py
@@ -96,7 +96,7 @@ class AvahiMonitor(object):
         """
 
         # XXX: Use the above native code for this.
-        check_call(
+        self.check_call(
             ["nfdc", "set-strategy", str(SUBSCRIPTIONS_SOMA), 'ndn:/localhost/nfd/strategy/multicast'])
 
         self.update_routed_status()
@@ -192,23 +192,23 @@ class AvahiMonitor(object):
                 "refusing to add %s as we'd go over %d peers", faceUri, MAX_PEERS)
             return
         self._peers.add(faceUri)
-        return check_call(["nfdc", "add-nexthop",
-                           "-c", str(defaults.RouteCost.LOCAL_NETWORK),
-                           str(SUBSCRIPTIONS_SOMA), faceUri])
+        return self.check_call(["nfdc", "add-nexthop",
+                                "-c", str(defaults.RouteCost.LOCAL_NETWORK),
+                                str(SUBSCRIPTIONS_SOMA), faceUri])
 
     def remove_nexthop(self, faceUri):
         try:
             self._peers.remove(faceUri)
-            check_call(["nfdc", "remove-nexthop",
-                        str(SUBSCRIPTIONS_SOMA), faceUri])
+            self.check_call(["nfdc", "remove-nexthop",
+                             str(SUBSCRIPTIONS_SOMA), faceUri])
         except KeyError:
             pass
 
     def start_edge(self):
-        return check_call(["systemctl", "start", "edd-soma-subscriptions-producer"])
+        return self.check_call(["systemctl", "start", "edd-soma-subscriptions-producer"])
 
     def stop_edge(self):
-        return check_call(["systemctl", "stop", "edd-soma-subscriptions-producer"])
+        return self.check_call(["systemctl", "stop", "edd-soma-subscriptions-producer"])
 
     def get_gateways(self):
         gateways = netifaces.gateways()
@@ -219,6 +219,12 @@ class AvahiMonitor(object):
         self.gateways = self.get_gateways()
         self.ips = flatten(
             map(lambda a: [x['addr'] for x in a], [flatten(netifaces.ifaddresses(i).values()) for i in netifaces.interfaces()]))
+
+    def check_call(self, a):
+        try:
+            return check_call(a)
+        except CalledProcessError as e:
+            logger.warning("Error calling: %s : %s", a, e)
 
 
 def main():
--- a/eos_data_distribution/avahi_monitor.py
+++ b/eos_data_distribution/avahi_monitor.py
@@ -69,6 +69,21 @@ class AvahiMonitor(object):
     """
 
     def __init__(self):
+        """
+        Attributes:
+            gateways (dict): Dictionary containing the network gateways
+                for this machine, as a mapping from address family to a
+                list of ``(address, interface, is_default)`` tuples.
+            _ndn_gateways (dict): Dictionary containing the addresses for
+                network gateways which advertise as NDN nodes, as a mapping
+                from key (see ``build_registry_key()``) to IP address.
+            _registry (dict): Dictionary containing all known NDN nodes
+                advertised over Avahi, as a mapping from key
+                (see ``build_registry_key()``) to face URI
+                (see ``face_uri_from_triplet()``).
+            _routed (bool): `True` if one of the gateways for this machine
+                is an NDN node; `False` otherwise.
+        """
         super(AvahiMonitor, self).__init__()
 
         self.ips = []
--- a/eos_data_distribution/avahi_monitor.py
+++ b/eos_data_distribution/avahi_monitor.py
@@ -187,10 +187,11 @@ class AvahiMonitor(object):
             logger.debug(
                 "refusing to add %s as we'd go over %d peers", faceUri, MAX_PEERS)
             return
-        self._peers.add(faceUri)
-        return self.check_call(["nfdc", "add-nexthop",
-                                "-c", str(defaults.RouteCost.LOCAL_NETWORK),
-                                str(SUBSCRIPTIONS_SOMA), faceUri])
+
+        if self.check_call(["nfdc", "add-nexthop",
+                            "-c", str(defaults.RouteCost.LOCAL_NETWORK),
+                            str(SUBSCRIPTIONS_SOMA), faceUri]):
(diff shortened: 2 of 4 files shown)


debian/control
```diff
@@ -2,21 +2,21 @@ Source: eos-data-distribution
...
+Build-Depends: debhelper (>= 10),
+               dh-python,
+               dh-systemd,
+               gir1.2-notify-0.7,
+               gir1.2-soup-2.4,
+               python-all,
+               python-avahi,
+               python-dbus,
+               python-docutils,
+               python-netifaces,
```
pwithnall: The reformatting here is unnecessary if you’re just adding `python-netifaces`.

debian/control
```diff
@@ -26,16 +26,15 @@ Vcs-Browser: https://github.com/endlessm/endless-ndn
...
- ${misc:Depends},
- ${python:Depends},
- adduser,
- nfd,
- gir1.2-notify-0.7,
- gir1.2-soup-2.4,
- python-avahi,
- python-dbus,
- python-pyndn,
+Depends: adduser,
```
pwithnall: Please don’t run `wrap-and-sort` apart from in a commit on its own. I used `wrap-and-sort -abst` on this packaging already.

eos_data_distribution/avahi_monitor.py
```diff
@@ -79,23 +96,102 @@ def __init__(self):
...
+                # this is limited to MAX_PEERS in add_nexthop
+                self._registry.values().map(self.add_nexthop)
+            else:
+                # nochange: wasn't routed, not routed now
+                pass
 
     def service_added_cb(self, sda, interface, protocol, name, type, h_type, domain, host, aprotocol, address, port, txt, flags):
-        ifname = sda.siocgifname(interface)
-        print "Found Service data for service '%s' of type '%s' (%s) in domain '%s' on %s.%i:" % (name, h_type, type, domain, ifname, protocol)
+        logger.debug(
```
pwithnall: Changing from `print` to `logger.debug` should ideally be done in a separate commit.

eos_data_distribution/avahi_monitor.py
```diff
@@ -59,13 +68,21 @@ class AvahiMonitor(object):
     def __init__(self):
         super(AvahiMonitor, self).__init__()
 
+        self.ips = []
```
pwithnall: This block of private variable declarations could do with some documentation on what they all mean and how they interact.

eos_data_distribution/avahi_monitor.py
```diff
@@ -79,23 +96,102 @@ def __init__(self):
...
+        self._routed = set(
+            self._ndn_gateways.values()).intersection(self.gateways)
+
+    def process_route_changes(self, prev_routed):
+        if self._routed:
+            logger.info(
+                "Being routed by an NDN node at %s (via %s), NOT acting as an edge router", address, ifname)
+            if not prev_routed:
+                # need to remove old links...
+                self._registery.values().map(self.remove_nexthop)
```
pwithnall: s/`self._registery`/`self._registry`/ (which suggests that this code path hasn’t been tested?)

eos_data_distribution/avahi_monitor.py
```diff
@@ -79,23 +96,102 @@ def __init__(self):
...
+    def add_nexthop(self, faceUri):
+        return check_call(["nfdc", "add-nexthop",
+                           "-c", str(defaults.RouteCost.LOCAL_NETWORK),
+                           str(SUBSCRIPTIONS_SOMA), faceUri])
+
+    def remove_nexthop(self, faceUri):
+        return check_call(["nfdc", "remove-nexthop",
+                           str(SUBSCRIPTIONS_SOMA), faceUri])
+
+    def refresh_network(self, *a, **kw):
```
pwithnall: Probably best to use the standard `*args, **kwargs` arguments for familiarity.

eos_data_distribution/avahi_monitor.py
```diff
@@ -79,23 +96,102 @@ def __init__(self):
...
+        self.gateways = netifaces.gateways()
```
pwithnall: As `self.gateways` has been updated here, you should update `self._routed` too.

eos_data_distribution/avahi_monitor.py
```diff
@@ -79,23 +96,102 @@ def __init__(self):
         check_call(
             ["nfdc", "set-strategy", str(SUBSCRIPTIONS_SOMA), 'ndn:/localhost/nfd/strategy/multicast'])
 
-        self._registry = dict()
+        self.update_routed_status()
+        self.process_route_changes(False)
+
+    def update_routed_status(self):
+        self._routed = set(
+            self._ndn_gateways.values()).intersection(self.gateways)
```
pwithnall: Will this ever set `self._routed` to `True`? If I understand correctly, `set(self._ndn_gateways.values())` is a set of IP address strings, and `self.gateways` is a dict of address family to a list of gateway information tuples*. How can they ever have an intersection?

* For example, from [here](https://pypi.python.org/pypi/netifaces#how-do-i-use-it): `{2: [('10.0.1.1', 'en0', True), ('10.2.1.1', 'en1', False)], 30: [('fe80::1', 'en0', True)], 'default': { 2: ('10.0.1.1', 'en0'), 30: ('fe80::1', 'en0') }}`

eos_data_distribution/avahi_monitor.py
```diff
@@ -79,23 +96,102 @@ def __init__(self):
...
+
+    def process_route_changes(self, prev_routed):
+        if self._routed:
+            logger.info(
+                "Being routed by an NDN node at %s (via %s), NOT acting as an edge router", address, ifname)
```
pwithnall: Where are `address` and `ifname` defined?

eos_data_distribution/avahi_monitor.py
```diff
@@ -79,23 +96,102 @@ def __init__(self):
...
+
         faceUri = face_uri_from_triplet(type, host, port)
-        check_call(["nfdc", "add-nexthop",
-                    "-c", str(defaults.RouteCost.LOCAL_NETWORK),
-                    str(SUBSCRIPTIONS_SOMA), faceUri])
-        self._registry[build_registry_key(name, type, domain)] = faceUri
+        key = build_registry_key(name, type, domain)
+        self._registry[key] = faceUri
+
+        if address in self.gateways:
```
pwithnall: I think this comparison will always be false, given the type of `self.gateways` is a dict mapping address family to a list of tuples.

eos_data_distribution/avahi_monitor.py
pwithnall: Is this called when any of this machine’s network interfaces go up or down?

eos_data_distribution/avahi_monitor.py
```diff
@@ -79,23 +96,102 @@ def __init__(self):
...
+            else:
+                # no change: i am routed, i was routed before: no links
+                pass
+        else:
+            logger.info(
+                "Not Being routed by an NDN node, acting as an edge router")
+            if prev_routed:
+                # need to add back old links...
+                # this is limited to MAX_PEERS in add_nexthop
+                self._registry.values().map(self.add_nexthop)
```
pwithnall: Do you need to remove the hops for the old `_ndn_gateways` here?

eos_data_distribution/avahi_monitor.py
```diff
@@ -79,23 +96,102 @@ def __init__(self):
...
+
+        if not self._routed and not prev_routed:
+            # not routed, wasn't routed, just remove this link
+            self.remove_nexthop(faceUri)
+        else:
+            # all the rest handled by generic code
+            self.process_route_changes()
+
+    def add_nexthop(self, faceUri):
+        return check_call(["nfdc", "add-nexthop",
```
pwithnall: These `check_call()` calls need some form of error handling. One failure from `nfdc` and this will raise a `CalledProcessError` which will wipe out the Avahi service.

eos_data_distribution/avahi_monitor.py
```diff
@@ -188,6 +192,12 @@ def remove_nexthop(self, faceUri):
         return check_call(["nfdc", "remove-nexthop",
                            str(SUBSCRIPTIONS_SOMA), faceUri])
 
+    def start_edge(self):
+        return check_call(["systemctl", "start", "edd-soma-subscriptions-producer"])
```
pwithnall: I don’t think this is the right architecture. This effectively means that the Avahi service runs to work out whether the `edd-soma-subscriptions-producer` can run — why not move the Avahi code into `edd-soma-subscriptions-producer`, so that the state information is kept within process, rather than being communicated between two processes via `systemctl` calls?

eos_data_distribution/avahi_monitor.py
```diff
@@ -33,6 +33,8 @@
 from eos_data_distribution.MDNS import ServiceDiscovery
 
 
+MAX_PEERS = 5
```
pwithnall: How was this number chosen? That should be added as a (docstring) comment by it.

eos_data_distribution/avahi_monitor.py
```diff
@@ -33,6 +33,8 @@
 from eos_data_distribution.MDNS import ServiceDiscovery
```
pwithnall: This commit is missing a commit message explaining why it’s a good idea to limit the number of peers.

eos_data_distribution/avahi_monitor.py
```diff
@@ -184,13 +187,22 @@ def service_removed_cb(self, sda, interface, protocol, name, type, domain, flags
             self.process_route_changes()
 
     def add_nexthop(self, faceUri):
+        if len(self._peers) > MAX_PEERS:
```
pwithnall: If the code ends up calling `add_nexthop()` 6 times (i.e. there are 6 valid next hops in the network), then `remove_nexthop()` for one of them once (for example, because that host disappears), we will end up with 4 peers registered in the RIB, not 5.

Taken to the limit, it would be possible for this machine to end up with zero hops in its RIB, even if several valid hops were present on the network, just because the first 5 hops to be added here were also the first (and only) 5 to be removed.

eos_data_distribution/avahi_monitor.py
```diff
@@ -184,13 +187,22 @@ def service_removed_cb(self, sda, interface, protocol, name, type, domain, flags
...
+            logger.debug(
+                "refusing to add %s as we'd go over %d peers", faceUri, MAX_PEERS)
+            return
+        self._peers.add(faceUri)
         return check_call(["nfdc", "add-nexthop",
                            "-c", str(defaults.RouteCost.LOCAL_NETWORK),
                            str(SUBSCRIPTIONS_SOMA), faceUri])
```
pwithnall: There is no binding between the success of the `nfdc add-nexthop` invocation and the `self._peers.add()` call here: the peer will always be added to `self._peers`, but then the `add-nexthop` invocation could fail, leaving the state out of sync.

eos_data_distribution/avahi_monitor.py
xaiki: erg... code moved around... fixing.

eos_data_distribution/avahi_monitor.py
xaiki: actually i do, but in removed_cb, because here if my disappeared host was a gateway it's not in `_ndn_gateways` anymore.
the other `_ndn_gateways` are either still there (but don't route us anymore) or we'll get called again to remove them.

eos_data_distribution/avahi_monitor.py
xaiki: nope, we trust avahi to handle that for us.

eos_data_distribution/avahi_monitor.py
xaiki: wrapped them.

eos_data_distribution/avahi_monitor.py
pwithnall: It would be clearer if the code were refactored to be more symmetric.

eos_data_distribution/avahi_monitor.py
pwithnall: Avahi will emit `service-added` or `service-removed` when one of the machine’s network interfaces goes up and down, even if there are no Avahi services on the subnet that interface is/was connected to?

xaiki: it will emit them when it goes down on an interface it was monitoring, that's what we're interested in. and when an interface comes up, if avahi handles it, it will start emiting `service-added`

so we're covered: we're only interested in interfaces avahi is monitoring anyway

eos_data_distribution/avahi_monitor.py
xaiki: very randomly, suggested by @ramcq

eos_data_distribution/avahi_monitor.py
xaiki: because that removes the hability to test edd-soma-subscriptions-producer on machines that do not have avahi.

eos_data_distribution/avahi_monitor.py
xaiki: well, there was because a failure in nfdc would kill the process.

eos_data_distribution/avahi_monitor.py
pwithnall: This is shipping on EOS: it will always have Avahi, right? In any case, the code here is essentially a policy for when to enable the SOMA producer. If we move this code into `edd-soma-subscriptions-producer` there is nothing stopping it also having other policies which could override this. Probably via a config file.

eos_data_distribution/avahi_monitor.py
xaiki: truedat, added a MIN_PEERS set to 2 so that we don't bother NFD too much too fast.

eos_data_distribution/avahi_monitor.py
xaiki: fixed

debian/control
xaiki: i don't care about the options we use, but they should be documented somewhere...

eos_data_distribution/avahi_monitor.py
xaiki: fixed

eos_data_distribution/avahi_monitor.py
xaiki: thought i answered, short answer: yes.

long answer: 
it will emit `service-removed` of all services on a iface it monitored that goes down.
it will emit `service-added` for all services on a iface it monitors when it goes up.

we do not care about ifaces avahi doesn't monitor because we have no way of finding peers there anyway.

debian/control
xaiki: fixed, please document the options you use with wrap-and-sort somewhere...

eos_data_distribution/avahi_monitor.py
xaiki: done

eos_data_distribution/avahi_monitor.py
xaiki: fixd

eos_data_distribution/avahi_monitor.py
xaiki: fixed

eos_data_distribution/avahi_monitor.py
xaiki: can't `self._ndn_gateways` still hasn't been updated, and i need `self.gateways` to update it.

eos_data_distribution/avahi_monitor.py
xaiki: well it used to die. fixed

eos_data_distribution/avahi_monitor.py
xaiki: because that prevents us to use `edd-soma-subscription-producer` in avahi-less machines, makes testability much harder.

eos_data_distribution/avahi_monitor.py
xaiki: introduced `MIN_PEERS` and a check for it so we don't hammer NFD.

eos_data_distribution/avahi_monitor.py
xaiki: pseudo randomly in conjunction with the @ramcq oracle.

eos_data_distribution/avahi_monitor.py
pwithnall: What Avahi-less machines? If `edd-soma-subscriptions-producer` had a config file allowing you to set the policy for when it would advertise SOMA subscriptions over NDN, the file could allow you to (for example) choose between `always`, `never` and `mdns-gateway` which would enable it based on the value of `self._routed` from this code.

xaiki: when it advertises is currently 'is it run or not'. i'd rather have edd-ssp do one thing (and do it well) and run it when i need it from the policy part.

pwithnall: There’s nothing stopping edd-ssp doing more than one thing well. Tying the policy of whether to advertise SOMA subscriptions with systemd’s run status for edd-ssp is conflating two things, which will result in confusion when one of them goes wrong. What if edd-ssp crashes at the same time as the policy decision from the Avahi service changes — systemd will unconditionally restart edd-ssp, but that might leave it running when the Avahi service doesn’t want it there. What if we add another component to endless-ndn which also has a say in the policy for running edd-ssp? Now both that component and the Avahi service need to communicate to work out whether to call `systemctl start` or `systemctl stop`. What if the initial call to `systemctl start` from the Avahi service fails? Should it start a timer and periodically retry starting edd-ssp? If so, that’s partially reinventing systemd’s job.

xaiki: crash: if you `stop` systemd will not try to restart. if it crashes, it's systemd's job to restart it.
`systemctl start` should not fail or something is wronger... you don't implement that in scripts that run it, you don't need to implement that in avahi.

about another service in NDN we'll solve it if/when we get there. it's the same problem anyway, you'd need to add that to ssp acording to your design, and i would like it even less.

eos_data_distribution/avahi_monitor.py
```diff
@@ -31,11 +32,19 @@
...
     # Disable TCP, we really only want UDP or ethernet
     # "_nfd._tcp",
     "_nfd._udp"]
 
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def flatten(l):
+    return [i for s in l for i in s]
```
wgreenberg: nit: I think `list(itertools.chain.from_iterable(l))` is cleaner, but no big deal

eos_data_distribution/avahi_monitor.py
xaiki: fxd

eos_data_distribution/avahi_monitor.py
```diff
@@ -76,26 +112,131 @@ def __init__(self):
...
+
+    def get_gateways(self):
+        gateways = netifaces.gateways()
+        del gateways['default']
+        return [g[0] for g in flatten(gateways.values())]
+
+    def refresh_network(self, *args, **kwargs):
+        self.gateways = self.get_gateways()
+        self.ips = flatten(
+            map(lambda a: [x['addr'] for x in a], [flatten(netifaces.ifaddresses(i).values()) for i in netifaces.interfaces()]))
```
wgreenberg: there's a lot going on here, IMO it'd be much clearer if split up over a few lines

eos_data_distribution/avahi_monitor.py
```diff
@@ -57,15 +69,39 @@ class AvahiMonitor(object):
     """
 
     def __init__(self):
+        """
+        Attributes:
+            gateways (dict): Dictionary containing the network gateways
```
wgreenberg: this seems to be out of date, since gateways is now a list

eos_data_distribution/avahi_monitor.py
```diff
@@ -57,15 +69,39 @@ class AvahiMonitor(object):
...
+                list of ``(address, interface, is_default)`` tuples.
+            _ndn_gateways (dict): Dictionary containing the addresses for
+                network gateways which advertise as NDN nodes, as a mapping
+                from key (see ``build_registry_key()``) to IP address.
+            _registry (dict): Dictionary containing all known NDN nodes
+                advertised over Avahi, as a mapping from key
+                (see ``build_registry_key()``) to face URI
+                (see ``face_uri_from_triplet()``).
+            _routed (bool): `True` if one of the gateways for this machine
+                is an NDN node; `False` otherwise.
```
wgreenberg: maybe add a comment for `ips` and `_peers` as well?

eos_data_distribution/avahi_monitor.py
xaiki: fxd

eos_data_distribution/avahi_monitor.py
xaiki: fxd

eos_data_distribution/avahi_monitor.py
xaiki: fxd

eos_data_distribution/avahi_monitor.py
xaiki: fxd, now it's readable for py-people and completly obscure for normal people.

wgreenberg:
CR pass!

PR #24, opened 2017-01-26
Next

xaiki:
tracking merges from next to master, i guess we can push this now, any objections @pwithnall ?


pwithnall:
> tracking merges from next to master, i guess we can push this now, any objections @pwithnall ?

No objections from me.

PR #25, opened 2017-03-13
eos-data-distribution (0.1.1) UNRELEASED; urgency=medium

xaiki:
* Split edd-avahi-monitor and edd-usb-producer to their own packages

 -- Niv Sardi <xaiki@endlessm.com>  Mon, 13 Mar 2017 11:26:48 -0300

https://phabricator.endlessm.com/T13940

Signed-off-by: Niv Sardi <xaiki@evilgiggle.com>

### DIFF
--- a/debian/rules
+++ b/debian/rules
@@ -8,6 +8,10 @@ export PYBUILD_NAME=eos-data-distribution
 
 override_dh_install:
 	dh_install --fail-missing
+	mkdir -p debian/edd-avahi-monitor/usr/bin
+	mv debian/python-eos-data-distribution/usr/bin/edd-avahi-monitor debian/edd-avahi-monitor/usr/bin
+	mkdir -p debian/edd-usb-producer/usr/bin
+	mv debian/python-eos-data-distribution/usr/bin/edd-usb-producer debian/edd-usb-producer/usr/bin
 
 # Build Sphinx documentation
 override_dh_sphinxdoc:
--- a/debian/rules
+++ b/debian/rules
@@ -1,17 +1,13 @@
 #! /usr/bin/make -f
 
 export PYBUILD_DISABLE=test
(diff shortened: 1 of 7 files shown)


debian/rules
```diff
@@ -8,6 +8,10 @@ export PYBUILD_NAME=eos-data-distribution
 
 override_dh_install:
 	dh_install --fail-missing
+	mkdir -p debian/edd-avahi-monitor/usr/bin
```
rshuler: @dbnicholson Is it typical to have to add this customization to debian/rules when splitting into multiple packages, or is there a better way?

dbnicholson: No, normally you would just put the files you want in the `$package.install` file and debhelper takes care of distributing them. In fact, I don't even know how the commands below would succeed. `dh_install` shouldn't be putting those files in the python-eos-data-distribution package since they're not listed in its install file.

The normal way to do this would be to put `usr/bin/edd-avah-monitor` in `edd-avahi-monitor.install` and `usr/bin/edd-usb-producer` in `edd-usb-producer.install`.

xaiki: the problem is that the python installer was putting these in the main package, `python-eos-data-distribution` any suggestions on how to fix that dh-python behaviour is welcome.

dbnicholson: Oh, it's because you're setting `PYBUILD_NAME` above. That breaks the multiple package setup in pybuild since it makes the destdir debian/python-eos-data-distribution instead of debian/tmp. Per pybuild(1), you just need to have debian/python-eos-data-distribution.install, which you have.

dbnicholson: I believe you could also set `PYBUILD_DESTDIR = debian/tmp`. In either case, you'd need to explicitly add the python files to the .install file.

xaiki: all fixed.

PR #26, opened 2017-04-20
Xaiki/no ndn

xaiki:
making a PR so we can discuss things here DO NOT MERGE


xaiki:
we're now at the point where demo.sh works, i think we could merge ?

PR #29, opened 2017-06-15
Xaiki/shortcut

xaiki:
This branch implements removing the store and assuming that a file is downloaded if found on disk

### DIFF
--- a/eos_data_distribution/ndn/dbus/base.py
+++ b/eos_data_distribution/ndn/dbus/base.py
@@ -237,6 +237,7 @@ class Consumer(Base):
             name, data = interface.call_request_interest_finish(res)
         except GLib.Error as error:
             # XXX actual error handeling !
+            logger.debug('got: %s, asuming TryAgain', error)
             # assuming TryAgain
             return self.expressInterest(interest)
 
@@ -278,8 +279,7 @@ class DBusProducerSingleton():
         object_skeleton.set_object_path(dbus_path)
         object_skeleton.add_interface(interface_skeleton)
 
-        logger.debug('registering path: %s ↔ %s', dbus_path,
-                     interface_skeleton.get_object_path())
+        logger.debug('registering path: %s', dbus_path)
         registered = self._object_manager.export(object_skeleton) or True
         iface_str = interface_skeleton.get_info().name
 
@@ -324,6 +324,7 @@ class DBusProducerSingleton():
 
         logger.debug('handeling call %s for name=%s, args=%s, kwargs=%s',
                      handler_name, name, args, kwargs)
+        logger.debug('cb_registery: %s', self._cb_registery)
 
         prefix = find_longest_prefix_in_list(name, self._cb_registery[handler_name].keys())
         if prefix:
@@ -401,7 +402,7 @@ class Producer(Base):
         self._dbus.return_value(name, name.toString(), data)
 
     def _on_request_interest(self, name, skeleton):
-        logger.debug('producer: got interest for name %s', name)
+        logger.debug('producer: got interest for name %s ↔ %s', name, self.name)
         self.emit('interest', name, Interest(name), None, None, None)
         return True
 
--- a/eos_data_distribution/ndn/dbus/base.py
+++ b/eos_data_distribution/ndn/dbus/base.py
@@ -183,8 +183,9 @@ class Consumer(Base):
         self.flush_pending_interests()
 
     def flush_pending_interests(self):
-        logger.debug('processing pending interests: %s', self._pending_interests.keys())
-        for interest in self._pending_interests.keys():
+        pending_interests = self._pending_interests.keys()
+        logger.debug('processing pending interests: %s', pending_interests)
+        for interest in pending_interests:
             self._dbus_express_interest(interest)
 
 
--- a/eos_data_distribution/ndn/dbus/base.py
+++ b/eos_data_distribution/ndn/dbus/base.py
@@ -18,6 +18,7 @@
 # A copy of the GNU Lesser General Public License is in the file COPYING.
 
 import logging
+import operator
 import gi
 
 gi.require_version('EosDataDistributionDbus', '0')
@@ -42,6 +43,11 @@ DBUS_PATH_TEMPLATE = '%s%s'
 dbus_producer_instances = dict()
 
 GObject.threads_init()
+
+def identity(*args):
+    if len(args) == 1:
+        return args[0]
+    return args
 
 def extend(a, b):
     a.extend(b)
@@ -70,6 +76,37 @@ def build_dbus_name(base, name):
     if component in ['soma', 'installed']:
         return base + '.' + component
     return base
+
+def find_longest_prefix_in_list(name, l, transform=identity):
+    found = None
+    prefix = name
+
+    def look_up(prefix, compare):
+        t = transform(prefix)
+        logger.debug("looking for %s in %s, with %s", t, l, compare)
+        try:
+            found = [k for k in l if compare(k, t)][0]
+            logger.debug("FOUND handler for %s ↔ %s", found, t)
+            return found
+        except IndexError:
+            logger.debug("Couldn't find handler for %s", t)
+
+        return False
+
+    found = look_up(prefix, str.startswith)
+    if found:
+        return found
+
+    prefix = '/'.join(prefix.split('/')[:-1])
+    while len(prefix):
+        found = look_up(prefix, operator.eq)
+        if found:
+            return found
+
+        prefix = '/'.join(prefix.split('/')[:-1])
+
+    logger.debug("FOUND NO handler for %s", name)
+    return None
 
 class Base(GObject.GObject):
     """Base class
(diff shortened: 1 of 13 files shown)


eos_data_distribution/ndn/file.py
```diff
@@ -186,21 +186,71 @@ def _on_file_monitor_changed(self, monitor, file, other_file, event_type):
...
+        # If we have an existing download to resume, use that. Otherwise,
+        # request the first segment to bootstrap us.
+        try:
+            self._read_segments_from_file(name)
+        except ValueError as e:
+            pass
+
+    def setName(self, name):
+        logger.debug('set name: %s ↔ %s', name, self.name)
+        if str(name) == str(self.name):
```
cosimoc: This looks like it could be useful in the superclass too

eos_data_distribution/ndn/file.py
```diff
@@ -186,21 +186,71 @@ def _on_file_monitor_changed(self, monitor, file, other_file, event_type):
...
         try:
-            # we need to make the dir early, so that the sgt file can be created
-            mkdir_p(os.path.dirname(filename))
-            self._segments_file = SegmentsFile(self._filename)
-            self._segments = self._segments_file.read()
+            self._read_segments_from_file(name)
         except ValueError as e:
             pass
 
+    def _read_segments_from_file(self, name):
```
cosimoc: You should not need to pass the name again here -- you can use `self.name`

eos_data_distribution/ndn/file.py
```diff
@@ -186,21 +186,71 @@ def _on_file_monitor_changed(self, monitor, file, other_file, event_type):
...
+        if not self._filename:
```
cosimoc: Looks like this should go in the constructor instead

eos_data_distribution/ndn/http.py
```diff
@@ -156,9 +156,10 @@ def _consume_queue(self):
 
 class Producer(chunks.Producer):
 
-    def __init__(self, name, url, session=None, *args, **kwargs):
-        self._getter = Getter(url, session=session,
+    def __init__(self, name, url, getter=None, session=None, *args, **kwargs):
+        getter = getter or Getter(url, session=session,
                               onData=lambda d: self.sendFinish(d))
```
cosimoc: Indentation a bit odd

eos_data_distribution/subscription.py
```diff
@@ -95,7 +96,7 @@ def _fetch_manifest_complete(self, consumer):
     def _on_shards_complete(self, parallel_consumer):
         response = {
             "subscription_id": self.subscription_id,
-            "manifest_path": self._manifest_filename,
+            "manifest_path": self._manifest_consumer._filename,
```
cosimoc: Does not look like `self._manifest_consumer` exists at this point

eos_data_distribution/subscription.py
```diff
@@ -62,14 +62,15 @@ def start(self):
...
         manifest_consumer = FileConsumer(
-            manifest_ndn_name, self._manifest_filename)
+            manifest_ndn_name, dirname=self._manifest_dirname)
         manifest_consumer.connect('complete', self._fetch_manifest_complete)
         manifest_consumer.start()
 
     def _fetch_manifest_complete(self, consumer):
-        with open(self._manifest_filename, 'r') as f:
+        # we can't know the filename in advance, as it's timestamped
+        with open(consumer._filename, 'r') as f:
```
cosimoc: Would be nice to have a getter instead of accessing private consumer data

eos_data_distribution/ndn/dbus/base.py
```diff
@@ -251,7 +251,9 @@ def __init__(self, name, dbus_name, skeleton):
         self._dbus_name = build_dbus_name(dbus_name, name)
         self._interface_skeleton_object = skeleton
 
-        self.con = Gio.bus_get_sync(BUS_TYPE, None)
+        address = Gio.dbus_address_get_for_bus_sync(BUS_TYPE, None)
+        self.con = Gio.DBusConnection.new_for_address_sync(address, Gio.DBusConnectionFlags.NONE, None, None)
```
cosimoc: This is really odd! Do things not work without it? I'd expect the address to return exactly the same value every time.

eos_data_distribution/ndn/dbus/base.py
```diff
@@ -251,9 +251,9 @@ def __init__(self, name, dbus_name, skeleton):
         self._dbus_name = build_dbus_name(dbus_name, name)
         self._interface_skeleton_object = skeleton
 
-        address = Gio.dbus_address_get_for_bus_sync(BUS_TYPE, None)
-        self.con = Gio.DBusConnection.new_for_address_sync(address, Gio.DBusConnectionFlags.NONE, None, None)
 
+        address = Gio.dbus_address_get_for_bus_sync(BUS_TYPE, None)
+        self.con = Gio.DBusConnection.new_for_address_sync(address, Gio.DBusConnectionFlags.AUTHENTICATION_CLIENT | Gio.DBusConnectionFlags.MESSAGE_BUS_CONNECTION, None, None)
```
cosimoc: Maybe squash with previous? (Even though it's still unclear why it's needed.)

eos_data_distribution/ndn/dbus/base.py
```diff
@@ -199,7 +199,13 @@ def expressInterest(self, interest=None, try_again=False):
         if not interest:
             interest = self.name.toString()
 
-        self.interest = interest
```
cosimoc, line 156: Is there nothing that relies on `self.interest` to be set?

eos_data_distribution/ndn/dbus/base.py
```diff
@@ -184,9 +184,9 @@ def _on_manager_ready(self, proxy, res):
 
     def flush_pending_interests(self):
         logger.debug('processing pending interests: %s', self._pending_interests.keys())
-        for interest, dbus_path, dbus_name in self._pending_interests.values():
-             if self._dbus_express_interest(interest, dbus_path, dbus_name):
-                del self._pending_interests[interest]
+        for interest in self._pending_interests.keys():
```
cosimoc: You could get the keys beforehand, since you use it twice in this function (or move the debug inside the loop)

eos_data_distribution/ndn/dbus/base.py
```diff
@@ -244,9 +244,13 @@ def _on_request_interest_complete(self, interface, res, interest):
             logger.debug('request-interest got error: %s', error)
             if str(error).find('ETRYAGAIN') != -1:
                 return self._dbus_express_interest(interest)
+            elif error.code == 24: # Timeout
```
cosimoc: Can you use `error.matches()` instead? That would allow disambiguating the error domain as well (the code alone is not enough to uniquely identify the error)

cosimoc (changes requested):
Thanks, @xaiki. Left some comments here.

cosimoc:
(Make sure to also look at the comments GitHub thinks are outdated, because not all of them may be.)

eos_data_distribution/ndn/http.py
xaiki: fixd

eos_data_distribution/subscription.py
xaiki: fixd

eos_data_distribution/ndn/dbus/base.py
xaiki: no it's an old API, we now support many interests (for re-entrance)

eos_data_distribution/ndn/dbus/base.py
xaiki: we used to need 2 DBus connections because if i reused the same one to publish on 2 different interfaces everything got exported to both.
this code is not strictly required now, but it's more correct than the former one.

eos_data_distribution/ndn/dbus/base.py
xaiki: fixd

eos_data_distribution/ndn/dbus/base.py
xaiki: from the docs: 
```
The returned object is a singleton, that is, shared with other callers of g_bus_get() and g_bus_get_sync() for bus_type . In the event that you need a private message bus connection, use g_dbus_address_get_for_bus_sync() and g_dbus_connection_new_for_address().
```

in store we have the base connection and we need another connection for the chunks code so that they are properly separated

xaiki: as for squashing... a lot of things must be squashed in these commits =/

eos_data_distribution/ndn/dbus/base.py
xaiki: i'm very much not sure how to use that API...
```
In [19]: int(Gio.DBusError.SPAWN_FORK_FAILED)
Out[19]: 24
```
that said when the error gets in i get:
```
GLib.Error: g-io-error-quark: Timeout was reached (24)
```

eos_data_distribution/ndn/file.py
xaiki: not like this, because it reads the segment file that makes no sense for a generic File.Consumer (opposed to a File.FileConsumer)

eos_data_distribution/ndn/file.py
xaiki: fixd

eos_data_distribution/ndn/file.py
xaiki: how so ? the trick is that we re-use it when ever the name changes this is why it was split out

xaiki: fixd

eos_data_distribution/ndn/dbus/chunks.py
```diff
@@ -201,15 +222,17 @@ def _on_progress(self, proxy, name, first_segment, last_segment):
...
+    def _complete(self):
+        if not self._emitted_complete:
             self._emitted_complete = True
-            proxy.call_complete(name, callback=self._on_complete)
+            self._interface.call_complete(
+                str(self.name), callback=self._on_complete)
 
     def _request_interest_complete(self, interface, res, interest):
         logger.info('Consumer: request interest complete: %s, %s, %s', interface, res, interest)
         name, final_segment, fd_list = interface.call_request_interest_finish(res)
```
cosimoc: Since you have removed the calls to `setName()` and `_set_final_segment()` here and we don't do anything with the return values, should something be changed in the dbus interface itself? It strikes me as odd.

xaiki:
(waiting for merge as we talked on today's 1:1)

eos_data_distribution/ndn/dbus/chunks.py
xaiki: it's a bit strange indeed, but it's left to have the same API as Base

